PR #6, opened 2014-12-08
Allow forcing a search to always run regardless of if the value exists or not

dakota:
Sometimes you need a finder or callback to run regardless of if there is a query value or not. This allows forcing a search to always run.

### DIFF
--- a/Type/Base.php
+++ b/Type/Base.php
@@ -56,7 +56,8 @@ abstract class Base {
 		$defaults = [
 			'field' => $name,
 			'name' => $name,
-			'validate' => []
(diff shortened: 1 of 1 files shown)


davidspeijer:
It is better to use if you want to reuse the code for for numeric values; int 0 is seen as empty!
```
if (!isset($args['created'])) {
	$args['created'] = new Time();
}
```

raul338:
@davidspeijer see [Empty fields](https://github.com/FriendsOfCake/search/tree/master/docs#empty-fields)

PR #14, opened 2015-05-18
Adding tests, some refactoring

burzum:
I've changed the namespace a little, I know it's a BC breaker but I felt it necessary.

Consider this PR as a RFC for now and provide feedback, thank you.


php-engineer:
:+1:

src/Search/Manager.php
```diff
@@ -112,7 +112,17 @@ public function callback($name, $config = [])
      */
     public function custom($name, $config = [])
     {
-        $this->_config[$name] = new $config['className']($name, $config, $this);
-        return $this;
+        if (isset($config['className'])) {
+            $this->_config[$name] = new $config['className']($name, $config, $this);
+               return $this;
```
jippi: indent too deep ?

src/Search/Type/Base.php
```diff
@@ -1,9 +1,9 @@
 <?php
-namespace Search\Type;
+namespace Search\Seach\Type;
 
 use Cake\ORM\Query;
 use Cake\Core\InstanceConfigTrait;
-use Search\Manager;
+use Search\Search\Manager;
```
jippi: why the double namespace Search ?

dereuromark: Well, it's the Search folder in the Search plugin, thus the double namespace is valid.
the vendor prefix would just be on top :)

burzum: I agree that it looks like a "double" but @dereuromark explained it right, it's because of the vendor + sub-folder namespace. Having the search related classes simply thrown into the "root" of the plugin doesn't feel right either.

src/Search/Manager.php
burzum: I'll fix that, happened probably because of some search and replace for tabs. God, I hate PSR2 so much...

src/Search/Type/Base.php
jippi: lets use `FOC\Search` then

like https://github.com/FriendsOfCake/Authenticate/ does for cake3 :)

src/Search/Manager.php
burzum: Fixed.

src/Search/Type/Base.php
burzum: OK, but you're aware this will become `namespace FOC\Search\Search` for the search related classes then?

burzum:
Tests can be run now, we just need to write tests. ;)

davidyell:
What benefits does having the extra namespace bring? I can't see any advantages.

jadb:
@burzum - I don't have the log for the vote on using FOC or not for BootstrapUI but the result was to **not** use it. The ones I am sure voted against are @lorenzo, and @ADmad, and if IIRC, @AD7six and @savant also. On another note, CRUD (and it's siblings, *flagship* of FOC, don't use it either). Let's just stick to 'Search' and not break namespace.

tests/Fixture/ArticlesFixture.php
```diff
@@ -0,0 +1,36 @@
...
+    ];
+
+/**
+ * Records
+ *
+ * @var array
+ */
+    public $records = [
+        [
+            'id' => '1',
```
ADmad: Postgres doesn't like `id` being specified.

tests/bootstrap.php
```diff
@@ -0,0 +1,53 @@
...
+unset($findRoot);
+chdir($root);
+if (file_exists($root . '/config/bootstrap.php')) {
+    require $root . '/config/bootstrap.php';
+    return;
+}
+require $root . '/vendor/cakephp/cakephp/tests/bootstrap.php';
+$loader = require $root . '/vendor/autoload.php';
+
+$loader->setPsr4('Cake\\', './vendor/cakephp/cakephp/src');
```
ADmad: These 2 statements aren't necessary.

ADmad:
@burzum Can you add a `.travis.yml` too so that builds get running.

tests/Fixture/ArticlesFixture.php
burzum: So what's the alternative?

burzum:
@ADmad yes I can do that.

tests/Fixture/ArticlesFixture.php
ADmad: Just remove it. `id` field will be auto filled.

jadb:
?

tests/Fixture/ArticlesFixture.php
burzum: I'm getting errors when I remove it. No matter if I remove it from the record array or the schema declaration.

burzum:
So when I change the namespace again any chance this is going to be accepted any  time soon, like in a few days? This goes to slow forward.

jadb:
I am :+1: for this once changes are done, but can't merge before at least someone else also approves it too.

burzum:
@jippi & @davidyell I've taken the license from one of my plugins but I've already removed this line?

See https://github.com/burzum/cakephp-search-plugin/blob/refactor/LICENSE.txt

ADmad:
I am not happy with change of file locations / namespace. I see no benefit of adding the extra `Search` folder. It leads to having `Search\Search\Foo` which looks weird.

dereuromark:
There is no additional folder. It is just a simple plugin + folder, as with all CakePHP 3 plugins so far.
It already dragged along too long, we should rather focus on shipping stuff like this before it gets discontinued for no good reason.

ADmad:
@dereuromark https://github.com/FriendsOfCake/search/blob/master/Type/Base.php vs https://github.com/burzum/cakephp-search-plugin/blob/refactor/src/Search/Type/Base.php

dereuromark:
I see, not only got src added but also Search. I bet the 2nd is an accident and is not how other plugins are done.

    src/Search/Type/Base.php => src/Type/Base.php

jippi:
+1 on flattening the directory hiearchy

davidyell:
I'm concerned that none of my questions have been addressed.

burzum:
Guys feel free to take the changes and finish it, I'm already using my fork in my project and sick of changing working things over and over.

The \Search\Search\ namespace was intentionally as for me it works like this:

```
<vendor>\<plugin>\<classTypeOrCategory>
FOC\Search\Search
```

For me this is far more logical than throwing classes directly into the root of the plugin like src/SomeClass.php. The plugin structure in the past always re-assembled the structure of a CakePHP app, so when had we ever files in in app\SomeClass.php?

PR #18, opened 2015-06-23
Prg

ADmad:


jippi:
tests be sad

ADmad:
@jippi Only phpcs build is failing because it doesn't like the order of params of `Base::_construct()`

jippi:
@ADmad possibruuu to fix?

ADmad:
Yeah i can swap the params keeping BC in a separate PR.

PR #22, opened 2015-06-30
Allowing nesting search query under 'search' key.

ADmad:
This avoid collision with other search options like 'group' etc.

### DIFF
--- a/src/Model/Behavior/SearchBehavior.php
+++ b/src/Model/Behavior/SearchBehavior.php
@@ -32,6 +32,10 @@ class SearchBehavior extends Behavior
      */
     public function findSearch(Query $query, array $options)
     {
+        if (isset($options['search'])) {
+            $options = $options['search'];
+        }
+
         foreach ($this->_table->searchConfiguration()->all() as $config) {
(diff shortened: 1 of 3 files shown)


src/Model/Behavior/SearchBehavior.php
```diff
@@ -45,7 +49,7 @@ public function findSearch(Query $query, array $options)
      * Returns the valid search parameter values according to those that are defined
      * in the searchConfiguration() method of the table.
      *
-     * @param array $param a key value list of search parameters to use for a search.
+     * @param array $params a key value list of search parameters to use for a search.
      * @return array
      */
     public function filterParams($params)
```
lorenzo: Should this function return the params nested in the the searchKey?

ADmad: If using the `search` key in options you wouldn't need to use this function.

ADmad: But will do so for consistency.

ADmad: Done

PR #29, opened 2015-08-07
Add Search form helper

### LINKED ISSUE #27: Add a helper to generate search form
I wrote a helper that automagically creates a complete search form based on the table's `searchConfiguration()` fields. I think it would be useful to include in this plugin.

``` php
use Cake\View\Helper;
use Cake\ORM\TableRegistry;

/**
 * SearchForm helper
 */
class SearchFormHelper extends Helper
{

    public $helpers = ['Form', 'Html'];

    /**
     * Default configuration.
     *
     * @var array
     */
    protected $_defaultConfig = [];

    /**
     * Output a complete search form, based on the search configurations set up
     * in the specified table
     *
     * @param $model string The Table to load
     * @return null|string An formatted FORM with inputs.
     */
    public function form($model)
    {
        $table = TableRegistry::get($model);

        // make sure the table has the behavior and has implemented the searchConfiguration method
        if (
            ! $table->behaviors()->has('Search') ||
            ! method_exists($table, 'searchConfiguration')
        ) {
            return null;
        }

        $searchConfig = $table->searchConfiguration()->all();

        $output = $this->Html->tag('h3', __('Search'));
        $output .= $this->Form->create();

        foreach ($searchConfig as $object) {
            $config = $object->config();

            // you can add stuff like `'multiple' => true` or `'empty' => ' '`
            // to the individual searchConfiguration field calls to be passed to
            // the form input
            $output .= $this->Form->input($config['name'], $config);
        }

        $output .= $this->Form->submit();
        $output .= $this->Form->end();

        return $output;
    }
}
```

The helper is loaded into the view, as normal, and is called in the template like this

``` php
<?= $this->SearchForm->form('Articles') ?>
```

(or if you want to be lazy, and copy-paste the same thing in all of your templates, you can call it like this)

``` php
<?= $this->SearchForm->form($this->request->params['controller']) ?>
```

search-form

curtisgibby:
Fix #27

### DIFF
--- a/src/View/Helper/SearchFormHelper.php
+++ b/src/View/Helper/SearchFormHelper.php
@@ -12,7 +12,7 @@ use Cake\ORM\TableRegistry;
 class SearchFormHelper extends Helper
 {
 
-    public $helpers = ['Form', 'Html'];
+    public $helpers = ['Form'];
 
     /**
      * Default configuration.
@@ -42,19 +42,20 @@ class SearchFormHelper extends Helper
 
         $searchConfig = $table->searchConfiguration()->all();
 
-        $output = $this->Html->tag('h3', __('Search'));
-        $output .= $this->Form->create();
+        $output = $this->Form->create();
 
+        $fields = [];
         foreach ($searchConfig as $object) {
             $config = $object->config();
 
             // you can add stuff like `'multiple' => true` or `'empty' => ' '`
             // to the individual searchConfiguration field calls to be passed to
             // the form input
-            $output .= $this->Form->input($config['name'], $config);
+            $fields[$config['name']] = $config;
         }
+        $output .= $this->Form->inputs($fields, ['legend' => __('Search')]);
 
-        $output .= $this->Form->submit();
+        $output .= $this->Form->button(__('Submit'));
         $output .= $this->Form->end();
 
         return $output;
--- a/src/View/Helper/SearchFormHelper.php
+++ b/src/View/Helper/SearchFormHelper.php
@@ -11,6 +11,11 @@ use Cake\View\View;
 class SearchFormHelper extends Helper
 {
 
+    /**
+     * Other helpers used by SearchFormHelper
+     *
+     * @var array
+     */
     public $helpers = ['Form'];
 
     /**
--- a/src/View/Helper/SearchFormHelper.php
+++ b/src/View/Helper/SearchFormHelper.php
@@ -1,0 +1,62 @@
+<?php
+namespace Search\View\Helper;
+
+use Cake\View\Helper;
+use Cake\View\View;
+use Cake\ORM\TableRegistry;
+
+
+/**
+ * SearchForm helper
+ */
+class SearchFormHelper extends Helper
+{
+
+    public $helpers = ['Form', 'Html'];
+
+    /**
+     * Default configuration.
+     *
+     * @var array
+     */
+    protected $_defaultConfig = [];
+
+    /**
+     * Output a complete search form, based on the search configurations set up
+     * in the specified table
+     *
+     * @param $model string The Table to load
+     * @return null|string An formatted FORM with inputs.
+     */
+    public function form($model)
+    {
+        $table = TableRegistry::get($model);
+
+        // make sure the table has the behavior and has implemented the searchConfiguration method
+        if (
+            ! $table->behaviors()->has('Search') ||
+            ! method_exists($table, 'searchConfiguration')
+        ) {
+            return null;
+        }
+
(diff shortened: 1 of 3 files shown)


src/View/Helper/SearchFormHelper.php
```diff
@@ -0,0 +1,62 @@
...
+        if (
+            ! $table->behaviors()->has('Search') ||
+            ! method_exists($table, 'searchConfiguration')
+        ) {
+            return null;
+        }
+
+        $searchConfig = $table->searchConfiguration()->all();
+
+        $output = $this->Html->tag('h3', __('Search'));
```
ADmad: Better to use `legend` tag of the form.

src/View/Helper/SearchFormHelper.php
```diff
@@ -0,0 +1,62 @@
...
+        foreach ($searchConfig as $object) {
+            $config = $object->config();
+
+            // you can add stuff like `'multiple' => true` or `'empty' => ' '`
+            // to the individual searchConfiguration field calls to be passed to
+            // the form input
+            $output .= $this->Form->input($config['name'], $config);
+        }
+
+        $output .= $this->Form->submit();
```
ADmad: Using `button()` allows better styling.

curtisgibby:
I've [updated](1a1cc341172df489c2aaf302b53a9946ecd9b8bf) the helper as suggested, using a form legend and a button.

src/View/Helper/SearchFormHelper.php
```diff
@@ -0,0 +1,62 @@
...
+        $fields = [];
+        foreach ($searchConfig as $object) {
+            $config = $object->config();
+
+            // you can add stuff like `'multiple' => true` or `'empty' => ' '`
+            // to the individual searchConfiguration field calls to be passed to
+            // the form input
+            $fields[$config['name']] = $config;
+        }
+        $output .= $this->Form->inputs($fields, ['legend' => __('Search')]);
```
ADmad: Use domain `search` for translated strings

curtisgibby: What about in the submit button's `__('Submit')`? There too?

curtisgibby:
Is there anything else you'd want done before this PR can be merged?

src/View/Helper/SearchFormHelper.php
```diff
@@ -0,0 +1,60 @@
...
+     *
+     * @param string $model The Table to load
+     * @return null|string An formatted FORM with inputs.
+     */
+    public function form($model)
+    {
+        $table = TableRegistry::get($model);
+
+        // make sure the table has the behavior and has implemented the searchConfiguration method
+        if (! $table->behaviors()->has('Search') ||
```
dereuromark: There should not be a space after `!`.

curtisgibby: Interesting that it passed the Travis / PHP-CS as-is... but I'll make the change.

src/View/Helper/SearchFormHelper.php
```diff
@@ -0,0 +1,60 @@
...
+use Cake\View\Helper;
+use Cake\View\View;
+
+/**
+ * SearchForm helper
+ */
+class SearchFormHelper extends Helper
+{
+
+    public $helpers = ['Form'];
```
ADmad: Docblock missing.

ADmad:
> Is there anything else you'd want done before this PR can be merged?

Tests :)

curtisgibby:
I was afraid that you were going to say that. :)

The helper is dependent on the Table having the Behavior. How do you mock that in a test?

ADmad:
Why mock it? Just attach the behavior to the table.

curtisgibby:
I don't have an `Articles` table in my project; I've got tables that have nothing to do with your tables. My tables do have the behavior, so I could write tests that would pass for me, but they'd fail for you. That's what I'm asking how to do: how to test a Table having a behavior when I can't guarantee that the table will exist in the testing environment.

ADmad:
Read the manual for how to use fixtures. The fixtures from are available to app/plugin tests too. So just create a table using one of the fixtures and attach behavior to that.

burzum:
I like the idea but the way it is implemented it seems to be not very flexible. Good for prototyping but not good enough for production. There is not enough control about the fields, the order of them and their look and feel.

The helper could use a template for the form in which it renders the fields into and maybe a callback that can be used on each field to add stuff before/after each input field or alter it.

ADmad:
Closing as this has been stalled for a long time. A new PR can be made taking into consideration all the feedback provided.

PR #32, opened 2015-09-02
Resolves #31

### LINKED ISSUE #31: Search.Prg killing request data
**What I did**
Added the `Search.Prg` component to my `AppController`.

**What I expected to happen**
I could save an edit form and the data would be saved and updated.

**What actually happened**
The form said it saved, yet no change to the data occurred.

**How I fixed it**
I removed the `Search.Prg` component and my form saved and updated as expected.

**What I think the problem is**
https://github.com/FriendsOfCake/search/blob/master/src/Controller/Component/PrgComponent.php#L20
An edit action is `put`, so this will wipe out the request data.

davidyell:
Removed the example where the component isn't wrapped in a check for an action.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,14 +78,6 @@ public function index()
 The `search` finder and the `filterParams()` method are dynamically provided by the
 `Search` behavior.
 
-* Then add the component search in the necessary methods (for our example index)
-```php
-public function index()
-{
-    $this->loadComponent('Search.Prg');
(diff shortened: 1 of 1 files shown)


ADmad:
In that example the component is loaded inside tge action itself.

davidyell:
Haha, one of those days I guess.

PR #34, opened 2015-09-16
Overhauled the docs. 

davidyell:
Organised the setup sections more logically. Added more comments to code sections to make specific things clearer. Added a note about the filterEmpty option. Added some more guidance on how to structure the setup

Tweaks to the comment length which were not wrapping nicely on Github. Also fixed minor typos and added some clarification.

A quick link to how it looks rendered, https://github.com/davidyell/search-1/blob/docs-update/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,50 +33,55 @@ or running command
 ```
 
 ## Usage
+The plugin has three main parts which you will need to configure and include in your application.
 
-* Add a search config method in your table
+### Table class
+There are three tasks during setup in your table class. Firstly you must add a `use` statement for the `Search\Manager`. Next you need to attach the Search behaviour to your table class. Lastly you must add a `searchConfiguration` method to your table class so that you can configure how the search will work.
+
 ```php
 use Search\Manager;
 
-...
+class ExampleTable extends Table {
 
-public function searchConfiguration()
-{
-    $search = new Manager($this);
-    $search
-    ->value('author_id', [
-        'field' => $this->aliasField('author_id')
-    ])
-    ->like('q', [
-        'before' => true,
-        'after' => true,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -84,15 +89,13 @@ public function index()
...
-public function index()
-{
-    $this->loadComponent('Search.Prg');
-}
-```
+### Component
+Then add the Search Prg component to the necessary methods in your controller. 
+
+:warning: Make sure, 
+* That you add this in the controllers `initialize` method.
```
ADmad: "controller's"

README.md
```diff
@@ -114,6 +119,21 @@ Would filter your list of articles to any article with "cakephp" in the `title`
 or `content` field. You might choose to make a `get` form which posts the filter
 directly to the url, or create links manually.
 
+### Creating your GET form
+You can change the method of a form when you create it.
+
+```php
+    echo $this->Form->create(null, ['type' => 'get']);
```
ADmad: If you use GET it will bypass the PRG process. Also you will end up with the token fields in query string if using security component. So the form should use POST.

jippi:
looks good!

README.md
davidyell: Ah ok, in the example I have running here, I'm using a get form and Prg and it's working as I expect. I'll update this then.

README.md
davidyell: Because the method belongs to the controller? I should probably learn this :cake:

bravo-kernel:
Nice job @davidyell and you beat me well to it, I can destroy the notes I took while finding this out by trial-and-error. Much obliged :relaxed:

davidyell:
Glad it's helpful! :+1:

PR #38, opened 2015-10-06
Fixes #30

### LINKED ISSUE #30: Remove csrfToken in PrgComponent
Do you think it would make sense to remove the `_csrfToken` in the `PrgComponent`? In general it's not necessary but it would keep the URL clean. What do you think? If you agree I will create a PR.

burzum:


steefaan:
:+1:

src/Controller/Component/PrgComponent.php
```diff
@@ -7,10 +7,22 @@
 class PrgComponent extends Component
 {
     /**
+     * $_defaultConfig For the Component.
+     *
+     * ### Options
+     * - `stripParams` Fields to strip from the data to query param conversion. Strips the `__csrfToken` by default.
```
ADmad: `__csrfToken` => `_csrfToken`

ADmad:
Ideally Cake itself should strip `_csrfToken` from `request->data`.

burzum:
@ADmad any reason this is still open? :(

ADmad:
cakephp/cakephp#7525 fixes the issue with csrf token. So is the change really necessary? I don't see the need to remove any other field to be common enough use case.

burzum:
OK, but then the minimum required CakePHP version of the search plugin should be raised to whatever version this fix is going to be in.

ADmad:
Yeah we can bump cake version as well as plugin's minor version after #40 is merged.

Issue #16, opened 2015-05-19
[RFC] new interface for searching


jippi:
(Originally posted by @burzum on the wiki)

Instead of relying on the presence of `searchConfiguration()` in the table class I think a trait that adds a `filters()` method that returns a manager instance would be better. The interface could be similar to validation:

```php

# Inside the Table instance

// Add filters to the "index" collection
$this->filters('index')
    ->add('field_name', 'like', $options)
    ->add('other_field', 'value', $options)

// Add filters to "otherName"
$this->filters('otherName')
    ->add('field_name', 'compare', $options);

// Returns all filters for the "index" collection
$this->filters('index')->get();

# Inside the Controller instance

// The find call could look like this then for example
 $query = $this->Country
    ->find('search', [$this->request->query, 'index'])
    ->where(['name !=' => null])
    ->order(['Country.id' => 'asc'])
    ->contain([
        'Cities'
    ]);

// Or maybe this?
 $query = $this->Country
    ->applyFilter('index', $this->request->query)
    ->where(['name !=' => null])
    ->order(['Country.id' => 'asc'])
    ->contain([
        'Cities'
    ]);
```

burzum:
@jippi since there is such an overwhelming interest from other people here is my implementation.

Please let me know if this will work for you and I'll add it to my branch.

```php
/**
 * Gets all filters in a given collection.
 *
 * @param string $collection Name of the filter collection.
 * @return array Array of filter instances.
 */
    public function getFilters($collection = 'default')
    {
        return $this->_filters[$collection];
    }

/**
 * Sets or gets the filter collection name.
 *
 * @param string $name Name of the active filter collection to set.
 * @return mixed Returns $this or the name of the active collection if no $name was provided.
 */
    public function collection($name = null) {
        if ($name === null) {
            return $this->_collection;
        }
        if (!isset($this->_filters[$name])) {
            $this->_filters[$name] = [];
        }
        $this->_collection = $name;
        return $this;
    }

/**
 * Adds a new filter to the active collection.
 *
 * @param string $name
 * @param string $filter
 * @param array $options
 * @return $this
 */
    public function add($name, $filter, array $options = [])
    {
        $this->_filters[$this->_collection][$name] = $this->_loadFilter($filter, $options);
        return $this;
    }

/**
 * Loads a search filter instance.
 *
 * @param string $name Name of the filter class to load.
 * @param array $options Filter options.
 * @return \Search\Search\Type\Base
[...]

PR #40, opened 2015-10-09
Fluid interface for the search manager.

burzum:
Implements https://github.com/FriendsOfCake/search/issues/16

### DIFF
--- a/src/Manager.php
+++ b/src/Manager.php
@@ -1,8 +1,9 @@
 <?php
 namespace Search;
 
-use Cake\Core\InstanceConfigTrait;
 use Cake\ORM\Table;
+use Cake\Utility\Inflector;
+use Search\Type;
 
 class Manager
 {
@@ -15,6 +16,22 @@ class Manager
     protected $_table;
 
     /**
+     * Filter collection and their filters
+     *
+     * @var array
+     */
+    protected $_filters = [
+        'default' => []
+    ];
+
+    /**
+     * Active filter collection.
+     *
+     * @var string
+     */
+    protected $_collection = 'default';
+
+    /**
      * Config
      *
      * @var array
@@ -24,7 +41,7 @@ class Manager
     /**
      * Constructor
      *
-     * @param \Cake\ORM\Table $table Table
+     * @param Table $table Table
      */
     public function __construct(Table $table)
     {
@@ -32,19 +49,19 @@ class Manager
     }
 
     /**
-     * Return all config
+     * Return all configured types.
      *
      * @return array Config
      */
     public function all()
     {
-        return $this->_config;
+        return $this->_filters['default'];
     }
 
     /**
      * Return Table
      *
-     * @return \Cake\ORM\Table Table Instance
+     * @return Table Table Instance
      */
     public function table()
     {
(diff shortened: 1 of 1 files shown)


src/Manager.php
```diff
@@ -52,80 +74,105 @@ public function table()
...
+        }
+        $typeClasses = $this->config('typeClasses');
+        if (isset($typeClasses[$filter])) {
+            return new $typeClasses[$filter]($filter, $this, $options);
+        }
+        if (class_exists('\Search\Type\\' . $filter)) {
+            $className = '\Search\Type\\' . $filter;
+            return new $className($name, $this, $options);
+        }
+        if (class_exists('\App\Search\Type\\' . $filter)) {
```
ADmad: It would be better to search for the class within App first and then in Search plugin.

ADmad:
@burzum Would it be possible to maintain backwards compatibility?

burzum:
@ADmad where does it break BC? All tests still pass.

src/Manager.php
burzum: But it's more likely you use filters from the plugin than your own?

ADmad:
The diff makes it look like methods were removed/changed but I guess it's misleading and method for adding filter are now handled by `__call()`.

That said existing test coverage is a bit slim so I wouldn't rely just on existing tests :)

burzum:
Well, add more tests. :stuck_out_tongue: I haven't changed the API, I'm just passing the calls through now.

src/Manager.php
ADmad: True. I was just thinking keeping it similar to core where you can transparently override a core class by using same named class within `App` namepsace. But it's pretty unlike someone would need to override the basic filter the plugin provides. So it's find to check in the plugin first.

ADmad:
> Well, add more tests.

It's on on long list of things to do :)

ADmad:
Tests are needed for the new methods that are added.

src/Manager.php
dakota: As it is, this will break if the app is not using the default `App` namespace. Any reason for this custom method over the built in CakePHP `App::className` method?

burzum: @dakota `$className = App::className($filter, 'Search/Type');` returns always false for me for some reason. I've debugged $fullname inside of className() and I've found `\Search\Type\Value`. I've tried a file_exists() on that myself and it returned true... No idea what's wrong. Try it:

```php
		$className = App::className($filter, 'Search/Type');
		if ($className) {
			return new $className($name, $this, $options);
		}
		throw new \InvalidArgumentException(sprintf('Can\'t find filter class "%s"!', $filter));
```

dakota: Looks like `App::className()` requires/expects you to use the plugin name, otherwise it will always search within the App namespace. `App::className($filter, 'Search/Type');` does not work, but `App::className('Search.' . $filter, 'Type');` does.

```php
$className = App::className($filter, 'Search/Type');
if (!$className) {
    $className = App::className('Search.' . str_replace('Search.', '', $filter), 'Type');
}
```

Should work as expected.

burzum: No, it doesn't for me, tried it as well.

PR #43, opened 2015-10-20
Updated the Like and Compare type to use a Comparison expression

Wouter0100:
This adds support for `ExpressionInterfaces` as field which, wasn't possible before due the string concate.
It uses a `Comparison`. The `columnType` is received from the `$field` when this `$field` is a `string`. When the `$field` isn't a `string`, we expect the `columnType` to be a `string` (defaults).

@Marlinc

### DIFF
--- a/tests/TestCase/Type/CompareTest.php
+++ b/tests/TestCase/Type/CompareTest.php
@@ -1,0 +1,47 @@
+<?php
+namespace Search\Test\TestCase\Type;
+
+use Cake\Core\Configure;
+use Cake\Database\Expression\BetweenExpression;
+use Cake\Database\Expression\FunctionExpression;
+use Cake\ORM\TableRegistry;
+use Cake\TestSuite\TestCase;
+use Search\Manager;
+use Search\Type\Compare;
+
+class CompareTest extends TestCase
+{
+
+    /**
+     * Fixtures
+     *
+     * @var array
+     */
+    public $fixtures = [
+        'plugin.Search.Articles'
+    ];
+
+    public function testProcess()
+    {
+        $articles = TableRegistry::get('Articles');
+        $manager = new Manager($articles);
+
+        $value = new Compare('created', $manager, [
+            'operator' => '>'
+        ]);
+        $value->args(['created' => '2012-12-12 12:12:12']);
+        $value->query($articles->find());
+        $value->process();
+
+        $this->assertEquals(1, $value->query()->count());
+
+        $value = new Compare('created', $manager, [
+            'operator' => '>='
+        ]);
+        $value->args(['created' => '2012-12-12 12:12:12']);
+        $value->query($articles->find());
+        $value->process();
+
+        $this->assertEquals(2, $value->query()->count());
+    }
+}
--- a/src/Model/Filter/Compare.php
+++ b/src/Model/Filter/Compare.php
@@ -1,6 +1,7 @@
 <?php
 namespace Search\Model\Filter;
 
+use Cake\Database\Expression\Comparison;
 use Cake\ORM\Query;
 
 class Compare extends Base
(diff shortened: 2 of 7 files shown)


src/Type/Compare.php
```diff
@@ -40,10 +41,14 @@ public function process()
...
-            $right = $this->value();
+            $columnType = 'string';
 
-            $conditions[] = [$left => $right];
+            if (is_string($field)) {
+                $columnExists = $this->manager()->table()->schema()->column($field);
+                $columnType = (!$columnExists) ? $this->manager()->table()->schema()->columnType($field) : 'string';
+            }
+
+            $conditions[] = new Comparison($field, $this->value(), $columnType, $this->config('operator'));;
```
dereuromark: two ;; at the end

Wouter0100: Whoops. I corrected it. Thx.

ADmad:
Tests needed

ADmad:
Looks like valid types for [$field](https://github.com/cakephp/cakephp/blob/master/src/Database/Expression/Comparison.php#L56) for `Comparison::__construct()` need to be updated.

src/Type/Compare.php
```diff
@@ -40,10 +41,14 @@ public function process()
...
         }
         foreach ($this->fields() as $field) {
-            $left = $field . ' ' . $this->config('operator');
-            $right = $this->value();
+            $columnType = 'string';
 
-            $conditions[] = [$left => $right];
+            if (is_string($field)) {
+                $columnExists = $this->manager()->table()->schema()->column($field);
+                $columnType = (!$columnExists) ? $this->manager()->table()->schema()->columnType($field) : 'string';
```
ADmad: Make a local var for `$this->manager()->table()->schema()` instead of using this long ass chain twice :)

Marlinc: @ADmad I'll take a look at this.

Wouter0100: @Marlinc Oops, sorry. I'm currently creating tests and also added a local variable for this :sweat_smile:

Wouter0100:
While creating tests for the `Compare` type, I came across a issue: I wanted to get the length of the title with the `FunctionExpression` and get all `Articles` with a title longer then 14 characters. This wasn't possible, because we assume a `FunctionExpression` is a `string` (which is the default).  What would the best way to set a `columnType`?

dereuromark:
This got stale but should be moved along IMO.

Marlinc:
Let me rebase the branch against master

Marlinc:
@dereuromark the changes have been rebased.

Is there something else that's needed for this?

Wouter0100:
Yes, @marlinc. Please see my response above... :(

tests/TestCase/Type/CompareTest.php
```diff
@@ -0,0 +1,47 @@
+<?php
+namespace Search\Test\TestCase\Type;
+
+use Cake\Core\Configure;
+use Cake\Database\Expression\BetweenExpression;
+use Cake\Database\Expression\FunctionExpression;
+use Cake\ORM\TableRegistry;
+use Cake\TestSuite\TestCase;
+use Search\Manager;
+use Search\Type\Compare;
```
josegonzalez: I believe this should be `Search\Model\Filter\Compare` perhaps.

ADmad:
@Wouter0100 @Marlinc Either of you want to finish up this PR? Based on @Wouter0100's last comment seems there are issues that still need to be addressed. We are going to make a new major release of the plugin so this would be a good time to wrap up this PR.

Marlinc:
I hope I might be able to work on this in a month or so.

dereuromark:
Yeah a minor after the next major sounds OK here.

jippi:
This PR seem to have stalled, closing

dereuromark:
Still valid, so should be opened as ticket then instead.

PR #44, opened 2015-11-01
Updated the callback's  reference object

damianoporta:
The name of the callback is `$this->config('name')` instead of `$this->config('callback')`, and it should point to the correct table, so i just added the reference of the table and the correct name with:

```
[$this->manager()->table(), $this->config('name')]
```

### DIFF
--- a/src/Type/Callback.php
+++ b/src/Type/Callback.php
@@ -17,6 +17,6 @@ class Callback extends Base
             return;
         }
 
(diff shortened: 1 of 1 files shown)


ADmad:
When using `Callback` type you are supposed to set option "callback" to a valid callable, it's not meant to just use the "name" as table method.

PR #46, opened 2015-11-04
Fluid manager

burzum:


ADmad:
Can you please add tests for `Manager::add()` with options and for custom `typeClasses` mapping.

burzum:
@ADmad done

tests/TestCase/ManagerTest.php
```diff
@@ -2,11 +2,17 @@
 namespace Search\Test\TestCase;
 
 use Cake\Core\Configure;
-use Cake\ORM\Entity;
-use Cake\ORM\Table;
 use Cake\ORM\TableRegistry;
 use Cake\TestSuite\TestCase;
 use Search\Manager;
+use Search\Type\base;
```
ADmad: Typo "base" => "Base"

burzum:
@ADmad fixed.

jippi:
-1 for killing existing method names - at least add them back for compat

burzum:
@jippi -1 for not reviewing the change properly. :stuck_out_tongue: 

```php
    public function __call($method, $args)
     {
        if (!isset($args[1])) {
            $args[1] = [];
        }
        return $this->add($args[0], $method, $args[1]);
     }
```

This will delegate the old method calls.

jippi:
got tests to prove it? :D

burzum:
The existing tests for two of the filters (why are they named types?) still pass. This is prove enough. I'm not going to add the tests you forgot to write initially.

jippi:
:dash:

ADmad:
@burzum Tests are missing for the new `Manager::collection()` method.

burzum:
@ADmad still something else...?

ADmad:
@burzum Looks good to me. Just one more thing. Even though the magic `__call()` takes of old methods it would be better to keep those concrete methods and make them use the new `add()` method.

The reasoning is similar to what @lorenzo stated in cakephp#7684. Having concrete method makes the API more readable and available methods easier to find.

burzum:
@ADmad and now...?

PR #48, opened 2015-11-14
Fluid manager

ADmad:
Refs #46

I simplified the `loadFilter()` method.

### DIFF
--- a/src/Manager.php
+++ b/src/Manager.php
@@ -1,8 +1,10 @@
 <?php
 namespace Search;
 
+use Cake\Core\App;
 use Cake\ORM\Table;
 use Cake\Utility\Inflector;
+use InvalidArgumentException;
 use Search\Type;
 
 class Manager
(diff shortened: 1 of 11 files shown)


jippi:
:+1:

src/Manager.php
```diff
@@ -125,7 +191,46 @@ public function compare($name, array $config = [])
...
+     */
+    public function loadFilter($name, $filter, array $options = [])
+    {
+        if (empty($options['className'])) {
+            $class = Inflector::classify($filter);
+        } else {
+            $class = $options['className'];
+            unset($options['className']);
+        }
+        $className = App::className($class, 'Type');
```
lorenzo: Hmm... having `src/Type/Like.php` would make absolutely no sense to me in my projects

ADmad: Yeah, that's a drawback of this simplification. We could change the namespace for the type classes. For one I would prefer renaming `Type` to `Filter`,  so that we have `src/Filter/Like.php` or better yet we can have `src/ORM/Filter/Like.php`.

lorenzo: ORM/Filter sounds good

ADmad: I actually meant `Model/Filter` :)  Will make the change.

ADmad:
@lorenzo Done

PR #49, opened 2015-11-16
ILIKE search method

urionlinejudge:
This commit enables to use ILIKE in SQL when necessary. For this we created a function following the LIKE statement in this plugin.

### DIFF
--- a/src/Model/Filter/ILike.php
+++ b/src/Model/Filter/ILike.php
@@ -1,0 +1,60 @@
+<?php
+namespace Search\Model\Filter;
+
+use Cake\ORM\Query;
+
+class ILike extends Base
+{
+
+    /**
+     * Default configuration.
+     *
+     * @var array
+     */
+    protected $_defaultConfig = [
+        'before' => false,
+        'after' => false,
+        'mode' => 'or'
+    ];
+
+    /**
+     * Process a ILIKE condition ($x ILIKE $y).
+     *
+     * @return void
+     */
+    public function process()
+    {
+        if ($this->skip()) {
+            return;
+        }
+
+        $conditions = [];
+        foreach ($this->fields() as $field) {
+            $left = $field . ' ILIKE';
+            $right = $this->_wildCards($this->value());
+
+            $conditions[] = [$left => $right];
+        }
+
+        $this->query()->andWhere([$this->config('mode') => $conditions]);
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


src/Model/Filter/ILike.php
```diff
@@ -0,0 +1,60 @@
+<?php
+namespace Search\Model\Filter;
+
+use Cake\ORM\Query;
+
+class ILike extends Base
```
ADmad: This class is a copy of `Like` with just a single line change. So instead just add a config to `Like` to allow using either `LIKE` or `ILIKE`.  Then `Manager::ilike()` call just use 'Search.Like' with appropriate config.

jippi: or make "Like" have a "comparisonKeyword" or something like that :p

ADmad: > or make "Like" have a "comparisonKeyword" or something like that :p 

That's what I meant by adding a config :)

burzum: +1 for making it a config of the existing like()

ADmad:
Closing due to lack of activity. You are welcome to reopen / make new PR as per changes suggested.

dereuromark:
When you close it you should (re)open a ticket for it, as it is actually missing functionality people should find when looking for it in the open issue section.

jippi:
i don't want stalled tickets laying around without anyone working on it

like everything else FoC, features and missing functionality is only done if someone cares enough - apparently no one cares enough about this feature, so don't see a need to keep the ticket or pr open :)

dereuromark:
Since when are u in charge?
Isnt this friends as in plural?

jippi:
I'm the goddamn boss of the intertubes! I'm the github santa and grinch..

lorenzo:
keep it friendly, guys :P

urionlinejudge:
Hi!

Sorry for not working on this sooner, I was a little busier than usual :)

I disagree that no one cares enough, because having support for ILIKE I believe to be more important than LIKE. Try finding usernames, post titles and other stuff without support for this... your search results would be incomplete in most cases.

Anyway, I made the suggested modifications and sent a pull request, please let me know if I can be o further assistance. =D

ADmad:
I share the same opinion as @jippi. If people care enough they should follow up on issues/PRs. Stalled issues will be cleaned up overtime.

PR #52, opened 2015-12-16
Made it easy to use PRG with just specific actions.

burzum:
It has been pretty annoying to repeat the action check over and over in the controllers. Let the component do this automatically for you now.

### DIFF
--- a/src/Controller/Component/PrgComponent.php
+++ b/src/Controller/Component/PrgComponent.php
@@ -9,6 +9,11 @@ class PrgComponent extends Component
     /**
      * Default config
      *
+     * ### Options
+     * - `actions` Method name(s) of actions to use PRG. Or bool for all or none.
+     *   You can pass a single action as string or multiple as array. Default is
+     *   true and all actions will be processed by the component.
+     *
      * @var array
      */
     protected $_defaultConfig = [
--- a/src/Controller/Component/PrgComponent.php
+++ b/src/Controller/Component/PrgComponent.php
@@ -28,9 +28,6 @@ class PrgComponent extends Component
      */
     public function startup()
     {
-        $this->controller = $this->_registry->getController();
-        $this->request = $this->controller->request;
-
         if ($this->_actionCheck()) {
             return $this->conversion();
         }
@@ -45,11 +42,13 @@ class PrgComponent extends Component
     public function conversion($redirect = true)
     {
         if (!$this->request->is('post')) {
-            $this->request->data = $this->controller->request->query;
+            $this->request->data = $this->request->query;
             return null;
         }
         if ($redirect) {
-            return $this->controller->redirect($this->request->params['pass'] + ['?' => $this->request->data]);
+            return $this->_registry->getController()->redirect(
+                $this->request->params['pass'] + ['?' => $this->request->data]
+            );
         }
         return null;
     }
--- a/src/Controller/Component/PrgComponent.php
+++ b/src/Controller/Component/PrgComponent.php
@@ -41,12 +41,12 @@ class PrgComponent extends Component
     {
         if (!$this->request->is('post')) {
             $this->request->data = $this->controller->request->query;
-            return;
+            return null;
         }
         if ($redirect) {
             return $this->controller->redirect($this->request->params['pass'] + ['?' => $this->request->data]);
         }
-        return;
+        return null;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Controller/Component/PrgComponent.php
```diff
@@ -2,26 +2,70 @@
 namespace Search\Controller\Component;
 
 use Cake\Controller\Component;
-use Cake\Event\Event;
 
 class PrgComponent extends Component
 {
+
+    /**
+     * Default config
```
davidyell: This doc block could do with some explanation in my opinion about the `actions` configuration setting. Due to it possibly containing `bool`, `string` or `array`. From reading the code I assume you configure which controller actions you want the PrgComponent to run on, in which case wouldn't `index` be the default as it's the most used use-case.

burzum: I'm not using `index` as default because of backward compatibility. If you don't do the check manually in the controller it will apply to all actions by default.

davidyell: If it applies to all actions, `edit` and `add` actions using post will have their form data removed, which I think will be confusing for beginners.

burzum: Yes, I agree with that. This "WTF?" happened to me as well when I started using the plugin. IMHO you should have to explicit set the actions instead of the component processing all by default.

src/Controller/Component/PrgComponent.php
```diff
@@ -2,26 +2,75 @@
...
      *
-     * @return void|Cake\Network\Response
+     * @return void|\Cake\Network\Response
      */
     public function startup()
     {
-        $controller = $this->_registry->getController();
-        $request = $controller->request;
+        $this->controller = $this->_registry->getController();
+        $this->request = $this->controller->request;
```
ADmad: `$this->request` is already set by component's constructor.

src/Controller/Component/PrgComponent.php
```diff
@@ -2,26 +2,75 @@
...
      * to the same action after converting the post data into GET params
      *
-     * @return void|Cake\Network\Response
+     * @return void|\Cake\Network\Response
      */
     public function startup()
     {
-        $controller = $this->_registry->getController();
-        $request = $controller->request;
+        $this->controller = $this->_registry->getController();
```
ADmad: The controller instance is redirected for using `Controller::redirect()` in `conversion()` so better to just use a local var in that method. I would prefer not adding extra properties.

src/Controller/Component/PrgComponent.php
```diff
@@ -2,26 +2,75 @@
...
+    /**
+     * POST to GET / GET to POST conversion
+     *
+     * @param bool $redirect Redirect on post, default true.
+     * @return \Cake\Network\Response|null
+     */
+    public function conversion($redirect = true)
+    {
+        if (!$this->request->is('post')) {
+            $this->request->data = $this->controller->request->query;
```
ADmad: `$this->controller->request->query` can be just `$this->request->query`.

burzum:
@ADmad anything else?

bravo-kernel:
Should this not be explained in the readme too?

netstyler:
:+1:

ADmad:
@burzum Looks good. Just the code eg. for component in readme should be updated to use the new config.

burzum:
@ADmad done.

src/Controller/Component/PrgComponent.php
```diff
@@ -2,26 +2,74 @@
...
+    protected $_defaultConfig = [
+        'actions' => true,
+    ];
+
     /**
      * Checks if the current request has posted data and redirects the users
      * to the same action after converting the post data into GET params
      *
-     * @return void|Cake\Network\Response
+     * @return void|\Cake\Network\Response
```
dereuromark: should also be `\Cake\Network\Response|null`

src/Controller/Component/PrgComponent.php
```diff
@@ -2,26 +2,74 @@
...
+    protected function _actionCheck()
+    {
+        $actions = $this->config('actions');
+        if (is_bool($actions)) {
+            return $actions;
+        }
+        if (is_string($actions)) {
+            $actions = [$actions];
+        }
+        if (in_array($this->request->action, $actions)) {
```
ADmad: Can just use `return in_array($this->request->action, $actions);`

bravo-kernel:
:clap:

Issue #53, opened 2015-12-21
Simplify the find('search') call


burzum:
Why do I have to explicitly call `filterParams()` here?

```php
->find('search', $this->Articles->filterParams($this->request->query))
```

I would move the call of `filterParams()` inside the behaviors [findSearch()](https://github.com/FriendsOfCake/search/blob/master/src/Model/Behavior/SearchBehavior.php#L46) method. If desired it can be still made configurable. This would remove the need to add the filterParams() call manually in the controller.

Another thing I don't think that is ideal is this inside findSearch():

```php
        if (isset($options['search'])) {
            $options = $options['search'];
        }
```

I would prefix the `search` with an `_` like the core does it with _serialize, _ext and a few others as well. Because `search` is a term that is likely to get used in a filter form.

dereuromark:
+100 on the second part. 

About the first, how do you pass in the query strings into the behavior then for the filterParams() call?

burzum:
@dereuromark well, simply without wrapping the query with the method? :)

```php
->find('search', $this->Articles->filterParams($this->request->query))
// vs
->find('search', $this->request->query)
```

dereuromark:
Ah, now I got it :)
I wonder if you ever needed the non filtered version. But it sounds reasonable to me to change it.

burzum:
> I wonder if you ever needed the non filtered (or custom filtered?) version.

Me too. I would always apply it by default and give the option to disable it for whatever possible reason.

davidyell:
These changes sound good to me :+1:

ADmad:
> ->find('search', $this->request->query)

Will cause bad things to happen if query string contains other valid find options like `group` etc.

ADmad:
Closing as related PR is open.

PR #54, opened 2015-12-21
Behavior improvement

burzum:


burzum:
Refs https://github.com/FriendsOfCake/search/issues/53

ADmad:
This change will break usages where users used `search` option directly like:

```php
find('search', ['search' => $this->request->query])
````

**Comment by ADmad** [comment]
Also as I commented on related issue, usage like
```php
    ->find('search', $this->request->query)
```
will cause bad things to happen if query string contains other valid find options like `group` etc.

I don't want to open the possibility of users shooting themselves in the foot.

josegonzalez:
@burzum friendly *bump* :)

burzum:
@josegonzalez  well, @ADmad basically rejected the idea. So just close it or change it in another way.

dereuromark:
We can do this now when we are fine with the 2.0.0 major bump.

dereuromark:
So should be merged along with https://github.com/FriendsOfCake/search/pull/83

src/Model/Behavior/SearchBehavior.php
```diff
@@ -45,8 +45,12 @@ class SearchBehavior extends Behavior
      */
     public function findSearch(Query $query, array $options)
     {
-        if (isset($options['search'])) {
-            $options = $options['search'];
+        if (!isset($options['_search']) && (!isset($options['_filter']) || $options['_filter'] === true)) {
+            $options = $this->filterParams($options);
```
ADmad: The check for `_search` and `_filter` should be mutually exclusive. `filterParams()` should be called after options are extracted from `$options['_search']` below.

burzum: Would you mind adding that change?

src/Model/Behavior/SearchBehavior.php
```diff
@@ -45,8 +45,12 @@ class SearchBehavior extends Behavior
      */
     public function findSearch(Query $query, array $options)
     {
-        if (isset($options['search'])) {
```
jippi: why would we use a `_` prefix ?

dereuromark:
So will this be included in the next major?

ADmad:
Yes, I am making an alternate PR.

burzum:
@ADmad why not just completing this one?

ADmad:
Because starting fresh myself was more convenient for me.

PR #58, opened 2015-12-23
ILIKE support

urionlinejudge:
Suggested modifications to support ILIKE.


jippi:
can you squash it down to one commit ? :) 

any maybe even add a small bonus test to it?

The change looks good

urionlinejudge:
Sure, I will add the test! But how can I "squash it down" to one commit?

jippi:
http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git

PR #64, opened 2016-01-12
Search by date range

aiphee:
Allow to search between date field like this:

```PHP
$search->date_between('publish_from');
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 vendor/
 phpunit.xml
 composer.lock
+.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 vendor/
 phpunit.xml
 composer.lock
-.idea/
--- a/src/Manager.php
+++ b/src/Manager.php
@@ -142,9 +142,19 @@ class Manager
         return $this;
     }
 
-    public function date_between($name, array $config = [])
+
+    /**
+     * Date between method
+     *
+     * @param string $name   Name
+     * @param array  $config Config
+     *
+     * @return $this
+     */
+    public function dateBetween($name, array $config = [])
     {
         $this->add($name, 'Search.DateBetween', $config);
+
         return $this;
     }
 
--- a/src/Manager.php
+++ b/src/Manager.php
@@ -142,6 +142,12 @@ class Manager
         return $this;
     }
 
+    public function date_between($name, array $config = [])
+    {
+        $this->add($name, 'Search.DateBetween', $config);
+        return $this;
+    }
+
     /**
      * finder method
      *
--- a/src/Manager.php
+++ b/src/Manager.php
@@ -154,7 +154,6 @@ class Manager
     public function dateBetween($name, array $config = [])
     {
         $this->add($name, 'Search.DateBetween', $config);
-
         return $this;
     }
 
--- a/tests/TestCase/Model/Filter/DateBetweenTest.php
+++ b/tests/TestCase/Model/Filter/DateBetweenTest.php
@@ -1,0 +1,43 @@
+<?php
+namespace Search\Test\TestCase\Model\Filter;
+
+use Cake\Core\Configure;
+use Cake\ORM\TableRegistry;
+use Cake\TestSuite\TestCase;
+use Search\Manager;
+use Search\Model\Filter\DateBetween;
+
+class DateBetweenTest extends TestCase
+{
+
+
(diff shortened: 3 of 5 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
 vendor/
 phpunit.xml
 composer.lock
+.idea/
```
ADmad: Don't add editor/environment specific ignores to repo. They should be in your system's global gitignore.

src/Model/Filter/Matching.php
```diff
@@ -0,0 +1,30 @@
+<?php
+	namespace Search\Model\Filter;
+
+	use Cake\I18n\Time;
+	use Cake\ORM\Query;
+
+	class Matching extends Base {
+
+		/**
+		 * Option to filter date range
```
ADmad: Incorrect

ADmad:
Tests are needed for the new filters

aiphee:
Ok, i have added tests, is this better?
This is my first pull -request, so sorry if it is messy.

tests/TestCase/Model/Filter/DateBetweenTest.php
```diff
@@ -0,0 +1,35 @@
...
+	use Search\Model\Filter\DateBetween;
+
+	class DateBetweenTest extends TestCase {
+
+
+		public $fixtures = [
+			'plugin.Search.Articles'
+		];
+
+		public function testProcess() {
```
jippi: can you make a test without `from` and `to` - and one where `to` are the only condition, and `last` the same

aiphee:
I just realized that im also committing matching condition, which i don't want to.

ADmad:
Use "git rm" to remove it

aiphee:
Alligned assignments like this:
```PHP
        $articles  = TableRegistry::get('Articles');
        $manager   = new Manager($articles);
        $valueFrom = $valueTo = new DateBetween('created', $manager);
```
are not allowed?

ADmad:
CakePHP's coding standards don't use aligned assignments. You can use https://github.com/cakephp/cakephp-codesniffer to run coding standards check locally.

Also please undo all new line and other changes to `Manager.php` and just add the new method. All the noise makes it difficult to review the patch.

aiphee:
Ok, phpcs got through, i reduced commits but i have no idea what this means in travis:
`Your requirements could not be resolved to an installable set of packages.`

src/Manager.php
```diff
@@ -142,6 +142,22 @@ public function value($name, array $config = [])
...
+     *
+     * @param string $name   Name
+     * @param array  $config Config
+     *
+     * @return $this
+     */
+    public function dateBetween($name, array $config = [])
+    {
+        $this->add($name, 'Search.DateBetween', $config);
+
```
jippi: no new-line here :)

aiphee:
No idea why the test are not passing, need to work on master branch, so i may create new pull request from other branch.

ealfonsov89:
i was working with the DateBetween filter and i find that the correct way to parse the date is `
            $date_from = new DateTime();
            $date_from->setDate($this->args()[$this->name()]['from']['year'],$this->args()[$this->name()]['from']['month'],$this->args()[$this->name()]['from']['day']);`, because thast's the default format used by the date form helper. the problem that i find now is that the filter does not execute, i presently start to use it and i not to mosh familiarize , but my configuration is:
 ReservsTable.php
`
        $this->addBehavior('Search.Search');
        $this->searchManager()
            ->date_between('date');`
ReservsController.php
`    public function initialize()
    {
        parent::initialize();
        $this->loadComponent('Search.Prg',['actions'=>['index']]);
    }
    /**
     * Index method
     *
     * @return \Cake\Http\Response|void
     */
    public function index()
    {
        $query = $this->Reservs->
            find('search',['search'=> $this->request->getQuery()]);
        $reservs = $this->paginate($query);
        $this->set(compact('reservs'));
        $this->set('_serialize', ['reservs']);
    }`
Index.ctp
`        <?= $this->Form->create(null, ['valueSources' => 'query'])?>
        <?= $this->Form->control('date.from',['type'=>'date'])?>
        <?= $this->Form->control('date.to',['type'=>'date'])?>
        <?= $this->Form->button('Filter', ['type' => 'submit','class'=>'button fa fa-filter'])?>
        <?= $this->Html->link('Reset', ['action' => 'index'],['class'=>'button fa fa-eraser'])?>`

Issue #67, opened 2016-01-27
Unable to redirect to named route


davidyell:
My project makes use of named routes in the front-end, and now I have a requirement to filter some data in the front-end of my project.

Unfortunately I hit a bug with the redirect, because it assumes that fallback routes are enabled, and doesn't specify the route.

https://github.com/FriendsOfCake/search/blob/master/src/Controller/Component/PrgComponent.php#L50

This will cause a Missing Route exception in projects which use named routes and do not have fallback routes enabled.

Also, just to compound matters I am using the same controller action on two routes to display two different sets of the same data. Open questions and closed questions.

```php
    $routes->connect(
        '/users/my-predictions',
        ['controller' => 'UserAnswers', 'action' => 'index', 'type' => 'open'],
        ['pass' => ['type'], '_name' => 'userOpenPredictions']
    );
    $routes->connect(
        '/users/past-predictions',
        ['controller' => 'UserAnswers', 'action' => 'index', 'type' => 'expired'],
        ['pass' => ['type'], '_name' => 'userPastPredictions']
    );
```

ADmad:
As you see that line of code just tries to redirect to current url with querystring. Your route has custom route element `type` which won't be included in the url array hence you don't get the expected result. 

I am not sure if there's a better way to guess the proper url. Only option I can think of is instead of array we use string url from `$this->request->url` and append the querystring. Currently you can use `beforeRedirect` callback to modify the url as required.

davidyell:
Ah yes, good idea. Thanks @ADmad

davidyell:
Closing in favour of #69

PR #69, opened 2016-01-29
Update to PrgComponent redirect

davidyell:
This update is to try and fix a bug detailed in #67 which I encountered when trying to redirect using named routes with no fallback routing.

I have updated the component to try and just 'copy' the existing url and append the query string, rather than trying to build a cake url array. This has proved successful in my specific use-case.

I've also updated the corresponding test case. I've had to make a few changes to it, such as putting the initialisation of the Component into the test case, so that the Request can be changed depending on the data from the newly created dataProvider method.

This should allow the test to expand further and easier in the future, if, for example, the test were to include some routing as well, the extra cases can be added.

I've also updated the `composer` file to include `phpunit` which seemed to be missing, as I needed to run the test suite.

### DIFF
--- a/src/Controller/Component/PrgComponent.php
+++ b/src/Controller/Component/PrgComponent.php
@@ -46,9 +46,10 @@ class PrgComponent extends Component
             return null;
         }
         if ($redirect) {
-            return $this->_registry->getController()->redirect(
-                $this->request->params['pass'] + ['?' => $this->request->data]
-            );
+            $parsedUrl = parse_url($this->request->here(false));
+            $url = $parsedUrl['path'] . '?' . http_build_query($this->request->data);
+
+            return $this->_registry->getController()->redirect($url);
         }
         return null;
     }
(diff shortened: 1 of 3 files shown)


src/Controller/Component/PrgComponent.php
```diff
@@ -46,9 +46,10 @@ public function conversion($redirect = true)
             return null;
         }
         if ($redirect) {
-            return $this->_registry->getController()->redirect(
-                $this->request->params['pass'] + ['?' => $this->request->data]
-            );
+            $parsedUrl = parse_url($this->request->here(false));
```
davidyell: I'm not 100% certain here, because `here()` does parse the `query` params inside the method, so it might be easier to set `Request::query` to the request data before calling `here()`. I would be interested in other people's thoughts on this approach.

ADmad: Parsing URL again to remove querystring could be avoided by using `'/' . $this->request->url` instead of `$this->request->here(false)` without would return same URL sans the querystring.

davidyell:
@ADmad Shall I close this PR? It seems your change supersedes this one.

ADmad:
Yes

PR #72, opened 2016-02-05
fallback to like

steefaan:
We have a postgres application but we also want to check against mysql with travis. If we use ILIKE, the test cases will fail for mysql. I think it would be nice to have an auto fallback to LIKE if ILIKE isn't available.

### DIFF
--- a/src/Model/Filter/Like.php
+++ b/src/Model/Filter/Like.php
@@ -1,6 +1,7 @@
 <?php
 namespace Search\Model\Filter;
 
+use Cake\Database\Driver\Postgres;
 use Cake\ORM\Query;
 
 class Like extends Base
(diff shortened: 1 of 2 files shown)


src/Model/Filter/Like.php
```diff
@@ -29,6 +30,11 @@ public function process()
             return;
         }
 
+        $driver = $this->query()->connection()->driver();
+        if (!$driver instanceof Postgres) {
```
josegonzalez: Maybe we should have a configurable map for the comparisons instead of doing this, that way we might accurately support all datastores?

steefaan: Can you give me an example how you would structure the `defaultConfig`?

ADmad:
I am not in favor of such automagic in the plugin.

In production no one really switches database so why test against db that you don't use?  Regardless, in your tests check the driver configured for `test` connection and accordingly switch the filter config.

josegonzalez:
An ILIKE isn't the same as LIKE in postgres, so that would also be a :-1: from me, since it might accidentally break functionality users would think would work, no?

PR #75, opened 2016-02-12
Update README.md

ionas:
... make it similar to FoC/cakephp-csvview

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,22 +12,24 @@ Search provides a search module for CakePHP applications.
 The master branch has the following requirements:
 
 * CakePHP 3.0.0 or greater.
+* Patience
 
 ## Installation
 
 * Install the plugin with composer from your CakePHP Project's ROOT directory
 (where composer.json file is located)
-```sh
-php composer.phar require friendsofcake/search
+```shell
+php composer.phar require friendsofcake/search:~1.2
 ```
 
-* Load the plugin by adding following to your `config/bootstrap.php`
-```php
-Plugin::load('Search');
-```
+### Enable plugin
 
-or running command
-```sh
+Load the plugin in your app's `config/bootstrap.php` file:
+
+	Plugin::load('Search');
+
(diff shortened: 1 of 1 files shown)


ionas:
Ref: https://github.com/FriendsOfCake/cakephp-csvview/pull/59

README.md
```diff
@@ -12,22 +12,24 @@ Search provides a search module for CakePHP applications.
 The master branch has the following requirements:
 
 * CakePHP 3.0.0 or greater.
+* Patience
```
ADmad: There is no need for this.

README.md
```diff
@@ -12,22 +12,24 @@ Search provides a search module for CakePHP applications.
...
 
 ## Installation
 
 * Install the plugin with composer from your CakePHP Project's ROOT directory
 (where composer.json file is located)
-```sh
-php composer.phar require friendsofcake/search
+```shell
+php composer.phar require friendsofcake/search:~1.2
```
ADmad: There's no need to specify version. Composer will automatically use latest release and set version constraints in composer.json

README.md
bravo-kernel: :flushed:

ionas:
In case you are happy with this, I will do the same for CsvView.

bravo-kernel:
I know it's just a matter of taste but I would suggest using the installation order as used in e.g. https://github.com/usemuffin/slug as that uses the IMHO preferable order to load the plugin (console first, manual otherwise). All to put new users onto the right path from the get go.

## Installation

Using [Composer](https://getcomposer.org)

```bash
composer require muffin/slug:~1.0
```

To make your application load the plugin either run:

```bash
./bin/cake plugin load Muffin/Slug
```

or add the following line to `config/bootstrap.php`:

```php
Plugin::load('Muffin/Slug');
```

davidyell:
I don't understand what benefits this brings, if any at all.

PR #79, opened 2016-03-01
Make search manager use RepositoryInterface instance instead of Table.

ADmad:
Refs UseMuffin/Webservice#18

### DIFF
--- a/src/Manager.php
+++ b/src/Manager.php
@@ -2,7 +2,7 @@
 namespace Search;
 
 use Cake\Core\App;
-use Cake\ORM\Table;
+use Cake\Datasource\RepositoryInterface;
 use Cake\Utility\Inflector;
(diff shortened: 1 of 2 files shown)


ADmad:
More changes will most likely be needed to make Search plugin work with non table datasource.

dereuromark:
Is this already good to go?

ADmad:
@jippi @lorenzo Any objections to merging this?

josegonzalez:
Note: behaviors cannot be used on non-Table objects since the typehint is `Table`, not `RepositoryInterface`. Changing would be a BC-break (I think?) and @lorenzo made it sound like he wants them to be a new type of thing [here](https://github.com/cakephp/elastic-search/issues/83#issuecomment-180597705).

PR #81, opened 2016-04-24
Fix query string building.

dereuromark:
Fixes the building of the url when query strings are not present

Currently

    /controller-name/search?

With the fix:

    /controller-name/search

### DIFF
--- a/src/Controller/Component/PrgComponent.php
+++ b/src/Controller/Component/PrgComponent.php
@@ -55,9 +55,11 @@ class PrgComponent extends Component
             }
 
             list($url) = explode('?', $this->request->here(false));
-            return $this->_registry->getController()->redirect(
-                $url . '?' . http_build_query($params)
-            );
+            $queryString = http_build_query($params);
+            if ($queryString !== '') {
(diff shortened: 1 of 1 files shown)


src/Controller/Component/PrgComponent.php
```diff
@@ -55,9 +55,11 @@ public function conversion($redirect = true)
             }
 
             list($url) = explode('?', $this->request->here(false));
-            return $this->_registry->getController()->redirect(
-                $url . '?' . http_build_query($params)
-            );
+            $queryString = http_build_query($params);
```
ADmad: Even the call to `http_build_query()` can be avoided by ensure `$params` is not empty.

dereuromark: That's probably true, but my main concern was the actual output.

ADmad: A separate PR isn't necessary for couple of lines of change :)

dereuromark: Then please feel free to add those on top. For me they are out of scope here.

Issue #82, opened 2016-04-26
Proper LIKE syntax for wildcards


dereuromark:
I would like to see this plugin also handle the LIKE syntax better than currently,
similar to the former cakedc one ( https://github.com/CakeDC/search/blob/master/Model/Behavior/SearchableBehavior.php#L31-L32 )

Users do not expect the sql syntax, nor should the plugin directly expose this sql specific one IMO.
Usually - like search engines do - they abstract it into a meta language, in this case using the well known `*` for any and `?` for a specific single char.

This is especially important as right now, if you need to look for those specific `%` or `_` chars, you get a lot of false positives here, e.g. when trying to find all that contain `22% discount` etc, as those are not escaped either and directly enter the DB search as wildcards.

And if you ever need `*` and `?` to be searchable (more unlikely as the other ones), then this would be configurable so you can adjust it on project level to be different ones.

I would like to know if there is an interest in implementing this in this plugin (only in the Like class of course).
If so I could look into fixing that in the near future.

jippi:
PR welcome

PR #83, opened 2016-04-28
Add char config for LIKE wildcards.

dereuromark:
Resolves https://github.com/FriendsOfCake/search/issues/82

Classified as bug as currently using wildcard chars as normal text triggers wildcard behavior of DB which is not intended (unless you configure and use those specifically as wildcard chars for your filter).

2nd bug was: When using array input, the LIKE before/after was not working.
There were also no tests for this bug, this is now also fixed.

Also some small code cleanup.

### DIFF
--- a/src/Controller/Component/PrgComponent.php
+++ b/src/Controller/Component/PrgComponent.php
@@ -74,6 +74,6 @@ class PrgComponent extends Component
         if (is_string($actions)) {
             $actions = [$actions];
         }
-        return in_array($this->request->action, $actions);
+        return in_array($this->request->action, $actions, true);
     }
 }
--- a/src/Model/Filter/Like.php
+++ b/src/Model/Filter/Like.php
@@ -14,7 +14,8 @@ class Like extends Base
         'after' => false,
         'mode' => 'or',
         'comparison' => 'LIKE',
-        'escapeWildcards' => false
+        'wildcardAny' => '*',
+        'wildcardOne' => '?',
     ];
 
     /**
@@ -31,9 +32,7 @@ class Like extends Base
         $conditions = [];
         foreach ($this->fields() as $field) {
             $left = $field . ' ' . $this->config('comparison');
-
-            $value = $this->_escapeWildCards($this->value());
-            $right = $this->_wildCards($value);
+            $right = $this->_wildCards($this->value());
 
             $conditions[] = [$left => $right];
         }
@@ -49,12 +48,20 @@ class Like extends Base
      */
     protected function _wildCards($value)
     {
+        if (is_array($value)) {
+            foreach ($value as $k => $v) {
+                $value[$k] = $this->_wildCards($v);
+            }
+            return $value;
+        }
+
+        $value = $this->_formatWildCards($value);
         if ($this->config('before')) {
-            $value = '%' . $value;
+            $value = $this->_formatWildCards($this->config('wildcardAny')) . $value;
         }
 
         if ($this->config('after')) {
-            $value = $value . '%';
+            $value = $value . $this->_formatWildCards($this->config('wildcardAny'));
         }
 
         return $value;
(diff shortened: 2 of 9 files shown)


ADmad:
Why all the unrelated changes in a single commit?

dereuromark:
Because I was working on a single branch for simplicity.
Those are simple fixes to easily look over and approve.

ADmad:
You can still do separate commits for unrelated changes on a single branch :)

dereuromark:
Or we can't be dense here and get stuff done ;)
You could just as well have fixed the issues I pointed out here yourself.

src/Model/Filter/Like.php
```diff
@@ -15,7 +15,9 @@ class Like extends Base
         'before' => false,
         'after' => false,
         'mode' => 'or',
-        'comparison' => 'LIKE'
+        'comparison' => 'LIKE',
+        'wildcardAny' => '*',
```
ADmad: If default value of `wildcardAny` is kept `*` it will require a major version bump for the plugin as it's a non BC change.

src/Controller/Component/PrgComponent.php
```diff
@@ -45,22 +45,23 @@ public function conversion($redirect = true)
             $this->request->data = $this->request->query;
             return null;
         }
-        if ($redirect) {
-            $params = $this->request->data;
+        if (!$redirect) {
+            return null;
```
jippi: why?

src/Controller/Component/PrgComponent.php
```diff
@@ -77,6 +78,6 @@ protected function _actionCheck()
         if (is_string($actions)) {
             $actions = [$actions];
         }
-        return in_array($this->request->action, $actions);
+        return in_array($this->request->action, $actions, true);
```
jippi: why safe?

src/Controller/Component/PrgComponent.php
ADmad: Early return, doesn't change actual behavior.

dereuromark: Its called return early and also called best practice :)

src/Controller/Component/PrgComponent.php
dereuromark: why not safe?!

src/Model/Filter/Like.php
dereuromark: True, are we fine with that? As I mentioned before no non-dev user would assume it to be % or _ anyway, they use the default wildcards from all OS and google and alike.

dereuromark:
Can we continue?

dereuromark:
Guys, whats the holdup?
I would very much not like to branch-off here as own composer plugin just because of that.
Let's please get things done.

dereuromark:
Closing and forking then as split off otherwise.. :/

src/Model/Filter/Like.php
josegonzalez: I'm fine with the bump.

src/Controller/Component/PrgComponent.php
josegonzalez: Fine by me :)

josegonzalez:
@dereuromark whats left on this PR, if anything?

dereuromark:
The decision on how to make a major bump probably.
And probably how to integrate https://github.com/FriendsOfCake/search/pull/85 before a major bump for all those that would need just escaping here.

dereuromark:
I rebased.
We can now release it with 2.0 tag and add documentation on how to use it from here on / what changed compared to 1.x.

src/Model/Filter/Like.php
```diff
@@ -14,7 +14,8 @@ class Like extends Base
         'after' => false,
         'mode' => 'or',
         'comparison' => 'LIKE',
-        'escapeWildcards' => false
```
ADmad: You removed the option to disable wildcard escaping?

dereuromark: Of course. Not necessary anymore.

dereuromark:
All good?

jippi:
TL;DR LGTM

ADmad:
TL;DR I'll make a 1.x release and then merge this to prepare for 2.0.

PR #90, opened 2016-06-03
Prg component

ADmad:
-  Run PRG conversion only for "index" action by default.
-  Allow disabling setting of querystring to request data.

### DIFF
--- a/src/Controller/Component/PrgComponent.php
+++ b/src/Controller/Component/PrgComponent.php
@@ -13,13 +13,13 @@ class PrgComponent extends Component
      * ### Options
      * - `actions` : Action name(s) to use PRG for. You can pass a single action
      *   as string or multiple as array. If boolean `true` all actions will be
-     *   processsed if `false` none. Default is ['index'].
+     *   processsed if `false` none. Default is ['index', 'lookup'].
      * - `queryStringToData` : Set query string as request data. Default `true`.
      *
      * @var array
      */
     protected $_defaultConfig = [
-        'actions' => ['index'],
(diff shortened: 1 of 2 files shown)


dereuromark:
I like it!

src/Controller/Component/PrgComponent.php
```diff
@@ -11,14 +11,16 @@ class PrgComponent extends Component
...
+     * - `actions` : Action name(s) to use PRG for. You can pass a single action
+     *   as string or multiple as array. If boolean `true` all actions will be
+     *   processsed if `false` none. Default is ['index'].
+     * - `queryStringToData` : Set query string as request data. Default `true`.
      *
      * @var array
      */
     protected $_defaultConfig = [
-        'actions' => true,
+        'actions' => ['index'],
```
jippi: will break a lot of crud-view :) the `lookup` action also uses search in all bownty apps

ADmad: This is gonna end up in a new major release. But adding `lookup` to the default list seems reasonable.

dereuromark: You can configure it, right?
It would be ideal if the configs would be merged with Configure::read('Search') to keep things DRY via configure.

ADmad: Yes `actions` is configurable`. I personally like to avoid reading too much stuff from `Configure`, you end up almost like using globals. Since `lookup` is a crud actions it's a sane default to have so that all FOC plugins inter-operate easily.

PR #91, opened 2016-06-03
Do away with SearchBehavior::filterParams().

ADmad:
Rename options key "search" to "_search" to allow using filter named "search".

Refs #54


ADmad:
A new major release can be done after this is merged. In spite of the changes I have maintained BC for calls like `find('search', $this->Posts->filterParams($this->request->query))`.

jippi:
👍

PR #92, opened 2016-06-05
Update readme.

ADmad:


davidyell:
Not that it matters, but I really don't like having so much configuration code in the `initialize()` method of my tables. I find it much easier from readability, discoverability and a code separation point of view to use the `searchConfiguration()` method. As it's nice and obvious.

So I really hope that, despite this change being merged, that the old method isn't removed. As I would be quite put off having, in some cases, huge chunks of method call inside a method of my Table class. When it could just be it's own method for clarity.

ADmad:
Support for `searchConfiguration()` hasn't been removed. I'll add back a short info in readme.

PR #93, opened 2016-06-07
Fixed small error in docs

ondrejhenek:
I'm not sure why the params were hidden under the '_search' key. Without it, the search works. If there is a reason for that, please, let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ public function index()
     $query = $this->Articles
         // Use the plugins 'search' custom finder and pass in the
         // processed query params
-        ->find('search', ['_search' => $this->request->query])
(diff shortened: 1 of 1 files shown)


ADmad:
1.x requires `search` key but current master code (and soon to be released 2.0.0) required `_search`.

ADmad:
Also while passing `$this->request->query` directly works you will have problems if query string contains param names like `group` which is a core find option.

ondrejhenek:
Thanks for the explanation. I'm going to use the `search` key then and wait for the 2.0.0. Cheers!

ADmad:
2.0.0 is already released and it still uses `search` key.

PR #95, opened 2016-06-07
Remove ability to directly pass querystring as find options.

ADmad:
I don't think it's prudent to allow usage like `find('search', $this->request->url)`. Passing querysting directly as find options is pretty risky thing imo. Keys could easily clash with core options or other custom options of custom finders. A user can for e.g. just tack on `&group=foo` and mess up your query. This patch will force people to always use `find('search', ['search' => $this->request->url])`.

I have also rename key "_search" back to "search" to maintain BC. Don't much value in adding the prefix and making people update their code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ public function index()
     $query = $this->Articles
         // Use the plugins 'search' custom finder and pass in the
         // processed query params
-        ->find('search', ['_search' => $this->request->query])
+        ->find('search', ['search' => $this->request->query])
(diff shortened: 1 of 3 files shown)


jippi:
LGTM

ADmad:
Will release 2.0.0 later today. It will still be pretty much BC and those using `find('search', $this->Table->filterParams($this->request->url))` shouldn't have any problems. The serarch listeners in other FOC plugins should also keep working without any change.

PR #98, opened 2016-06-08
Auto alias fields in filters.

ADmad:
This saves users using Table::aliasField() each time when configuring filters
to avoid "ambiguous column" sql errors.

I'll add / update tests if people like the idea.


dereuromark:
Do we need a test for this to avoid regressions?
Apart from that: looks good so far.

jippi:
a test or two would be lovely @ADmad 
the change looks legit

ADmad:
Tests added.

PR #99, opened 2016-06-08
Fixes to method comments

johanmeiring:
I picked up two inaccuracies in method comments of two of the filter classes.  This PR should fix that.  Hope it helps :)


ADmad:
Thanks.

PR #105, opened 2016-06-29
Resolves #104

### LINKED ISSUE #104: Like filter not using aliasField()
When using the `like` filter the fields are not being wrapped by `Base::field()` because they're instead using `Base::fields()` which doesn't call `field()` and thus the configuration for using `aliasField()` isn't respected.

This is contrary to the release notes, https://github.com/FriendsOfCake/search/releases/tag/2.0.1

This means that the following produce different results for my setup,

```php
// Ambiguous field error
        $search
            ->like('name', [
                'before' => true,
                'after' => true,
                'field' => 'name'
            ])

// Works as expected
        $search
            ->like('name', [
                'before' => true,
                'after' => true,
                'field' => $this->aliasField('name')
            ])
```

davidyell:
Instead of reading the field from the configuration use the method, so that the `aliasFields` setting is respected.

Updated the matching test as well.

### DIFF
--- a/src/Model/Filter/Base.php
+++ b/src/Model/Filter/Base.php
@@ -114,7 +114,7 @@ abstract class Base
      */
     public function fields()
     {
-        return (array)$this->config('field');
+        return (array)$this->field();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


ADmad:
Nice find!

PR #106, opened 2016-07-04
add @throws

thinkingmedia:


dereuromark:
Should be FQCN

thinkingmedia:
@dereuromark true, but it was already listed in the "use"

dereuromark:
Yeah, makes the doc block still not more valid :)

Issue #110, opened 2016-08-09
Add Search.Boolean


htstudios:
```php
<?php
namespace Search\Model\Filter;

class Boolean extends Base
{

    /**
     * Default configuration.
     *
     * @var array
     */
    protected $_defaultConfig = [
        'truthy' => [1, true, '1', 'true',  'yes', 'on'],
        'falsy' => [0, false, '0', 'false', 'no',  'off']
    ];

    /**
     * Check if a value is truthy/falsy and pass as condition.
     *
     * @return void
     */
    public function process()
    {
        if ($this->skip()) {
            return;
        }

        $value = strtolower($this->value());
        $bool = null;
        if (in_array($value, $this->config('truthy'), true)) {
            $bool = true;
        } elseif (in_array($value, $this->config('falsy'), true)) {
            $bool = false;
        }

        if ($bool !== null) {
            $this->query()->andWhere([$this->field() => $bool]);
        }
    }
}
```

~~Could also use FILTER_VALIDATE_BOOLEAN http://php.net/manual/de/filter.filters.validate.php... nopes, would return false for "".~~

~~... however through Session injection Search.Boolean could translate boolean representations coming from the user.~~

htstudios:
Because of https://github.com/FriendsOfCake/search/issues/111 at the current state this code snipplet works better:

```
<?php
namespace Search\Model\Filter;

class Boolean extends Base
{

    /**
     * Process a value condition ($x == $y).
     *
     * @return void
     */
    public function process()
    {
        if ($this->skip()) {
            return;
        }

        $falsy = ['0', 'false', 'no', 'off'];
        $truthy = ['1', 'true', 'yes', 'on'];

        $value = strtolower($this->value());
        $bool = null;
        if (in_array($value, $falsy, true)) {
            $bool = false;
        } elseif (in_array($value, $truthy, true)) {
            $bool = true;
        }

        if ($bool !== null) {
            $this->query()->andWhere([$this->config('field') => $bool]);
        }
    }
}

jippi:
can you please open a PR ? :)

htstudios:
Closing as PR is open.

PR #112, opened 2016-08-10
Added Search.Boolean

htstudios:
Ref: https://github.com/FriendsOfCake/search/issues/110

### DIFF
--- a/src/Model/Filter/Boolean.php
+++ b/src/Model/Filter/Boolean.php
@@ -1,0 +1,40 @@
+<?php
+namespace Search\Model\Filter;
+
+class Boolean extends Base
+{
+
+    /**
+     * Default configuration.
+     *
+     * @var array
+     */
+    protected $_defaultConfig = [
+        'truthy' => [1, true, '1', 'true', 'yes', 'on'],
+        'falsy' => [0, false, '0', 'false', 'no', 'off']
+    ];
+
+    /**
+     * Check if a value is truthy/falsy and pass as condition.
+     *
+     * @return void
+     */
+    public function process()
+    {
+        if ($this->skip()) {
+            return;
+        }
+
+        $value = strtolower($this->value());
+        $bool = null;
+        if (in_array($value, $this->config('truthy'), true)) {
+            $bool = true;
+        } elseif (in_array($value, $this->config('falsy'), true)) {
+            $bool = false;
+        }
+
+        if ($bool !== null) {
+            $this->query()->andWhere([$this->field() => $bool]);
+        }
+    }
+}
--- a/tests/Fixture/ArticlesFixture.php
+++ b/tests/Fixture/ArticlesFixture.php
@@ -14,6 +14,7 @@ class ArticlesFixture extends TestFixture
     public $fields = [
         'id' => ['type' => 'string', 'null' => false, 'length' => 36],
         'title' => ['type' => 'string', 'null' => false, 'default' => null],
+        'is_active' => ['type' => 'boolean', 'null' => false, 'default' => true],
         'created' => ['type' => 'datetime', 'null' => true, 'default' => null],
         'modified' => ['type' => 'datetime', 'null' => true, 'default' => null],
     ];
@@ -27,6 +28,7 @@ class ArticlesFixture extends TestFixture
         [
             'id' => '1',
             'title' => 'Test title one',
+            'is_active' => true,
             'created' => '2012-12-12 12:12:12',
             'modified' => '2013-01-01 11:11:11',
         ],
--- a/tests/TestCase/Model/Filter/BooleanTest.php
+++ b/tests/TestCase/Model/Filter/BooleanTest.php
@@ -1,0 +1,129 @@
+<?php
+namespace Search\Test\TestCase\Model\Filter;
+
+use Cake\Core\Configure;
+use Cake\ORM\Entity;
+use Cake\ORM\Table;
+use Cake\ORM\TableRegistry;
+use Cake\TestSuite\TestCase;
+use Search\Manager;
+use Search\Model\Filter\Base;
+use Search\Model\Filter\Boolean;
+
+class BooleanTest extends TestCase
+{
+
+    /**
+     * Fixtures
+     *
+     * @var array
+     */
+    public $fixtures = [
+        'plugin.Search.Articles'
+    ];
+
+    public function testProcessWithFlagOn()
+    {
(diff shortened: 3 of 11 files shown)


src/Model/Filter/Boolean.php
```diff
@@ -0,0 +1,40 @@
...
+class Boolean extends Base
+{
+
+    /**
+     * Default configuration.
+     *
+     * @var array
+     */
+    protected $_defaultConfig = [
+        'truthy' => [1, true, '1', 'true',  'yes', 'on'],
```
dereuromark: I would make those class const

    const TRUTHY = 'truthy';
    const FALSY = 'falsy';

Then one can use $this->config(static::TRUTHY) etc

jippi: doing that is not really consistent with how cake or the library in general does things though :)

jippi:
LGTM

src/Model/Filter/Boolean.php
dereuromark: I disagree :)
It kinda does - https://github.com/cakephp/cakephp/blob/master/src/Filesystem/Folder.php#L37 as only of of many - but it sure does way too less in some cases.
But you can also keep this way.

jippi: when the constant have the same name as the value, it makes little sense to me :) 

if it was `::POSITIVE = 'truthy';` and `::NEGATIVE = 'falsely';` it could make sense - let's keep it as is

dereuromark: It makes sense either way - only then you could for example refactor without danger of forgetting things. But never mind that now.

All right then.

dereuromark:
:+1:

src/Model/Filter/Boolean.php
```diff
@@ -0,0 +1,40 @@
...
+    {
+        if ($this->skip()) {
+            return;
+        }
+
+        $value = strtolower($this->value());
+        $bool = null;
+        if (in_array($value, $this->config('truthy'), true)) {
+            $bool = true;
+        } elseif (in_array($value, $this->config('falsy'), true)) {
```
ADmad: There is no test coverage for falsy case.

tests/Fixture/ArticlesFixture.php
```diff
@@ -13,9 +13,10 @@ class ArticlesFixture extends TestFixture
      */
     public $fields = [
         'id' => ['type' => 'string', 'null' => false, 'length' => 36],
-        'title' => ['type' => 'string', 'null' => false, 'default' => null],
         'created' => ['type' => 'datetime', 'null' => true, 'default' => null],
         'modified' => ['type' => 'datetime', 'null' => true, 'default' => null],
+        'title' => ['type' => 'string', 'null' => false, 'default' => null],
```
ADmad: Why create noise by unnecessarily moving this line? :)

ADmad:
CS [error](https://travis-ci.org/FriendsOfCake/search/jobs/151215548#L278)

src/Model/Filter/Boolean.php
htstudios: I am not opposed on using consts to get closer to atoms (Elixir, Ruby, Python). However CakePHP is often not doing that.

It is something I'd like to see for 4.x or 5.x ... e.g. moving to only constants and methods and never hash assignments of options, as its ultra error prone and thus consumes **MEGATIME** (sorry ;p)

tests/Fixture/ArticlesFixture.php
htstudios: Cause my brain imploded when seeing the un-cakeish order of fields ;p
I can undo that.

src/Model/Filter/Boolean.php
jippi: lets keep it like this

src/Model/Filter/Boolean.php
jippi: would be nice to have :)

tests/Fixture/ArticlesFixture.php
jippi: for me its cake to have created/modified last :P

ADmad: Yip i follow the same

htstudios: I think it would be worthwhile to at least recommend one way in `cakephp/docs` (and I don't care which way, it just makes it easier to work on cake apps in future). Reverted.

htstudios:
@ADmad took that as a pledge to fix all the other CS issues as well (lets see what CS Checker says).

src/Model/Filter/Boolean.php
htstudios: done.

ADmad:
> @ADmad took that as a pledge to fix all the other CS issues as well

I already fixed CS error in other files on master branch but thanks :)

composer.json
```diff
@@ -19,7 +19,9 @@
     ],
     "require-dev": {
         "cakephp/cakephp": "~3.0",
-        "friendsofcake/cakephp-test-utilities": "dev-master"
+        "friendsofcake/cakephp-test-utilities": "dev-master",
+        "cakephp/cakephp-codesniffer": "2.*",
+        "phpunit/phpunit": "^5.5"
```
ADmad: Please don't change the composer file. You can install codesniffer globally locally. Also phpunit 5.5 requires PHP 5.6+ and we need to test on 5.4 and. So just undo changes to composer.json

tests/TestCase/Model/Filter/BooleanTest.php
```diff
@@ -0,0 +1,130 @@
...
+        $boolean->args(['is_active' => 'neitherTruthyNorFalsy']);
+        $boolean->query($articles->find());
+
+        $query = $boolean->query();
+        $this->assertEmpty($query->clause('where'));
+
+        $boolean->process();
+        $this->assertEmpty($query->clause('where'));
+    }
+
```
ADmad: https://travis-ci.org/FriendsOfCake/search/jobs/151520329#L221

dereuromark: The opening brace should also not have newlines before content.

htstudios:
Yay Travis went :green:!

jippi:
@htstudios can you include update to README.md with the filter type as well?

htstudios:
Done.

jippi:
Thanks @htstudios !

PR #113, opened 2016-08-10
allow prefixed/scoped/nested/dotnotated search params

htstudios:
However the UnitTest is broken as somehow the `searchManager()->add` calls seem to change nothing. Any hints?


davidyell:
I don't understand the point of this feature. To me you could achieve the same using a `finder` type search, and then use a custom finder to contain any related data you wanted.

htstudios:
As explained in IRC, you cannot. The point is that you do not have to write a ton of `finder` types and still be able to use multiple filters on the same single view (say /users/view/1).

jippi:
there seem to be failing tests :)

htstudios:
### Unit tests
I understood the source of the problem. 
Had to create a new test and fixture to avoid class name clashes and setup the manager.
Fixed it - but I am quite unhappy how I did it. Any ideas or is it fine?

htstudios:
The failing test seems to be unrelated?!

ADmad:
https://travis-ci.org/FriendsOfCake/search/jobs/151476411#L220

ADmad:
Why did you make a new test case file instead of just adding new tests to existing `SearchBehaviorTest` ?

htstudios:
@ADmad because in my process of working I had a lot of issues with the custom class ahead the Test File. I can probably merge stuff back now and simply use the comments fixture, ok?

ADmad:
> I can probably merge stuff back now and simply use the comments fixture, ok?

Yes. You could also use a core fixture instead of making a new one in plugin. Core fixtures are available in [plugin tests](https://github.com/FriendsOfCake/search/blob/master/composer.json#L32).

htstudios:
I would have kept the tests files separate as every test now needs to load the comments table, comments search setup, comments fixture. Anyway I merged that.

I did also use the core.Comments fixture - that's a lot better, thank you for the pointers!

Any more concerns, else I will try to squash the commit chaos.

ADmad:
> every test now needs to load the comments table, 

You could have avoided that by loading table only in the required test method instead of `setUp()` but it doesn't really matter.

htstudios:
Thanks for the quick @jippi merges and all the help, especially by @ADmad !

PR #114, opened 2016-08-22
Example on how to persist query strings

inoas:
Example on how to persist query strings through filtering

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,7 +220,7 @@ echo $this->Form->input('author_id', ['empty' => 'Pick an author']);
 
 ## Persisting the Query String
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -217,3 +217,14 @@ Be sure to allow empty in your search form, if you're using one.
 ```php
 echo $this->Form->input('author_id', ['empty' => 'Pick an author']);
 ```
+
+## Persisting the Query String
```
dereuromark: query strings (plural) IMO.

inoas: As far as I know not. https://en.wikipedia.org/wiki/Query_string

dereuromark:
This example lacks the page omission.
Without page being removed it will break with NotFoundException s.

Issue #84, opened 2016-04-28
There should be a "isSearch" bool


dereuromark:
The CakeDC search has a bool flag "isSearch" to be [passed down to the view](https://github.com/CakeDC/search/blob/master/Controller/Component/PrgComponent.php#L250) so the form knows if a "reset" button or alike is needed.
This is currently not possible with this one.
You cannot just use !empty(querystring) because that could contain some order/page params etc.

Is there an idea on how to get this implemented?
Then I can take a look into this.

dereuromark:
feedback?

jippi:
PRG could append ?_isSearch=1 to the query when redirecting ? :)

dereuromark:
That is not always true though, as sometimes it does not apply filters (like clearing or just using defaults again). So this would have to be done only for get request and only if any of the behavior's query modifications occur.
also in general it would be best to just have an internal variable passed down to be used in the view layer etc to toggle then.
The question is now this can be done.

jippi:
considering we only got a model/table/behavior going for us, it would have to be through Configure to push the state, or up to the developer to just mark it on his own.

burzum:
> PRG could append ?_isSearch=1 to the query when redirecting ? :)

The PRG component or behavior could detect if any searchable query param was present and implement a method `isSearch` that returns bool.

htstudios:
It should take some kind of scope, if you have multiple filters per page.
The name should be configurable. By default something like `searching=true` or `searching[CountriesTable] = true` (latter with scopes).

dereuromark:
If the component has a method for this, we don't necessarily need the query string approach.
Anyone a concrete idea how to implement?

inoas:
How would you differentiate across different Cells?

dereuromark:
@inoas This works on a query basis on the model layer, cells are out of scope here, you will need to do that yourself somehow.

inoas:
I am not so sure its a good solution (as the boolean state that one really wants should store if a query/search had been executed, and as such that's hard to know from the model layer and should be in the controller layer where user interaction is handled best), but if it works for you guys, nvm.

PR #118, opened 2016-09-10
Allow controller to ask the model for isSearch().

dereuromark:
Resolves https://github.com/FriendsOfCake/search/issues/84
```php
$articles = $this->paginate($query);
debug($this->Articles->isSearch());

$this->set('isSearch', $this->Articles->isSearch()); 
```

### DIFF
--- a/src/Model/Behavior/SearchBehavior.php
+++ b/src/Model/Behavior/SearchBehavior.php
@@ -35,8 +35,12 @@ class SearchBehavior extends Behavior
             'searchManager' => 'searchManager',
             'isSearch' => 'isSearch'
         ],
-        'isSearch' => false
     ];
+    
+    /**
+     * @var bool
+     */
+    protected $_isSearch = false;
 
     /**
      * Callback fired from the controller.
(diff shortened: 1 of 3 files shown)


dereuromark:
We could make 'isSearch' just a protected $isSearch attribute on the behavior.
No need to expose it.

A cleaner approach would be to have each finder process() method return a bool.
True indicates that you actually modified the query, false would mean you bailed early, no modification done/needed after all.
Would that be a more desired solution here?

ADmad:
CS error.

src/Model/Behavior/SearchBehavior.php
```diff
@@ -32,9 +32,15 @@ class SearchBehavior extends Behavior
...
             'filterParams' => 'filterParams',
-            'searchManager' => 'searchManager'
-        ]
+            'searchManager' => 'searchManager',
+            'isSearch' => 'isSearch'
+        ],
     ];
+    
+    /**
+     * @var bool
```
ADmad: Description missing

dereuromark:
Also added docs.

Issue #117, opened 2016-09-07
SearchBehavior unable to use non default filters collection.


ADmad:
The `SearchManager` has the ability to maintain multiple filter collections. For e.g. you can have separate collections for backend and frontend. But when using the search finder there's no way to tell it to which collection to use. It always uses the "default" collection.

The fix seems easy enough. Just allow passing a `collection` key to finder options and update the behavior to use `SearchBehavior::getFilters()` passing in the collection name instead of calling `SearchBehavior::all()` which always returns the default collection.

flangofas:
I will provide a fix for this one.

ADmad:
@flangofas Great! Don't forget tests :)

PR #119, opened 2016-09-11
Allow multiple filter collections

flangofas:
Related to: https://github.com/FriendsOfCake/search/issues/117

### DIFF
--- a/src/Model/Behavior/SearchBehavior.php
+++ b/src/Model/Behavior/SearchBehavior.php
@@ -42,6 +42,7 @@ class SearchBehavior extends Behavior
      * @param Query $query Query.
      * @param array $options The options for the find.
      *   - `search`: Array of search arguments.
+     *   - `collection`: Filter collection name.
      * @return \Cake\ORM\Query The Query object used in pagination.
      * @throws \Exception When missing search arguments.
      */
@@ -53,8 +54,7 @@ class SearchBehavior extends Behavior
                 'to be nested under key "search" in find() options.'
             );
         }
-
-        $filters = $this->_getAllFilters();
+        $filters = $this->_getAllFilters(Hash::get($options, 'collection', 'default'));
         $params = (array)$options['search'];
         $params = Hash::flatten($params);
         $params = array_intersect_key(Hash::filter($params), $filters);
@@ -97,17 +97,20 @@ class SearchBehavior extends Behavior
     }
 
     /**
-     * Gets all filters from the search manager.
+     * Gets all filters by the default or given collection from the search manager
      *
+     * @param string|null $collection name of collection
      * @return \Search\Model\Filter\Base[] An array of filters for the defined fields.
      */
-    protected function _getAllFilters()
+    protected function _getAllFilters($collection = 'default')
     {
         $method = $this->config('searchConfigMethod');
         if (method_exists($this->_table, $method)) {
-            return $this->_table->{$method}()->all();
+            $manager = $this->_table->{$method}();
+        } else {
+            $manager = $this->searchManager();
         }
 
-        return $this->searchManager()->all();
+        return $manager->getFilters($collection);
     }
 }
--- a/tests/TestCase/Model/Behavior/SearchBehaviorTest.php
+++ b/tests/TestCase/Model/Behavior/SearchBehaviorTest.php
@@ -36,6 +36,21 @@ class CommentsTable extends Table
     }
 }
 
+class GroupsTable extends Table
+{
+
+    public function searchConfiguration()
+    {
+        $manager = new Manager($this);
+
+        return $manager
+            ->collection('frontend')
+            ->value('title')
+            ->collection('backend')
+            ->like('title', ['before' => true, 'after' => true]);
+    }
+}
+
 class SearchBehaviorTest extends TestCase
 {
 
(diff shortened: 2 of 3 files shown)


src/Model/Behavior/SearchBehavior.php
```diff
@@ -53,8 +54,12 @@ public function findSearch(Query $query, array $options)
                 'to be nested under key "search" in find() options.'
             );
         }
-
-        $filters = $this->_getAllFilters();
+        $collection = Hash::get($options, 'collection');
+        if (is_string($collection) && !empty($collection)) {
+            $filters = $this->getFilters($collection);
```
ADmad: A new method isn't necessary. You can just add the collection argument to `_getFilters()` method and use a one liner `$filters = $this->_getAllFilters(Hash::get($options, 'collection'));`

flangofas: @ADmad done.

src/Model/Behavior/SearchBehavior.php
```diff
@@ -97,17 +97,20 @@ public function searchManager()
     }
 
     /**
-     * Gets all filters from the search manager.
+     * Gets all filters by the default or given collection from the search manager
      *
+     * @param  string $collection name of collection
```
dereuromark: double space

src/Model/Behavior/SearchBehavior.php
```diff
@@ -97,17 +97,20 @@ public function searchManager()
...
         $method = $this->config('searchConfigMethod');
         if (method_exists($this->_table, $method)) {
-            return $this->_table->{$method}()->all();
+            $manager = $this->_table->{$method}();
+        } else {
+            $manager = $this->searchManager();
         }
 
-        return $this->searchManager()->all();
+        return !is_null($collection) ? $manager->getFilters($collection) : $manager->all();
```
dereuromark: !== null

flangofas: @dereuromark done.

src/Model/Behavior/SearchBehavior.php
flangofas: @dereuromark done.

src/Model/Behavior/SearchBehavior.php
```diff
@@ -111,6 +111,6 @@ protected function _getAllFilters($collection = null)
             $manager = $this->searchManager();
         }
 
-        return !is_null($collection) ? $manager->getFilters($collection) : $manager->all();
+        return ($collection !== null) ? $manager->getFilters($collection) : $manager->all();
```
inoas: Is this exactly the same in PHP?

dereuromark: It is.

inoas: Makes me wonder why `is_null()` was introduced to PHP then.
Thanks for the cleanups

flangofas:
any ideas why the tests are failing only in PHP 5.4 mysql and pgsql?

tests/TestCase/Model/Behavior/SearchBehaviorTest.php
```diff
@@ -36,6 +36,21 @@ public function searchConfiguration()
...
+class GroupsTable extends Table
+{
+
+    public function searchConfiguration()
+    {
+        $manager = new Manager($this);
+
+        return $manager
+            ->collection('frontend')
+            ->value('Groups.title')
```
ADmad: Remove the `Group.` alias. It's now done automatically in 2.x version of the plugin. The tests are failing because of that on PHP 5.4. They pass on 5.5+ in CakePHP 3.3 `aliasField()` was updated to ensure already aliased fields were not aliased again.

ADmad:
Does anyone see the custom find option key `collection` conflicting?  Should we perhaps name it `_collection` or `searchCollection` ?

tests/TestCase/Model/Behavior/SearchBehaviorTest.php
flangofas: @ADmad done.

dereuromark:
I is probably quite unlikely to be conflicting, And right now it would be consistent to search one.

ADmad:
@flangofas Can you please squash the commits into one.

flangofas:
@ADmad updated.
I think ```Manager::all()``` (https://github.com/FriendsOfCake/search/blob/master/src/Manager.php#L50) should support other collections as well and not only the default one. What do you think?

ADmad:
@flangofas `Manager::getFilters()` already does that so no point having 2 methods doing the same. IIRC `Manager::all()` was kept for backwards compatibility so let's keep it unchanged.

ADmad:
This is good to merge once you can squash the commits into one.

tests/TestCase/Model/Behavior/SearchBehaviorTest.php
```diff
@@ -36,6 +36,21 @@ public function searchConfiguration()
     }
 }
 
+class GroupsTable extends Table
```
dereuromark: Ideally, all those classes should be moved into the tests/TestApp/ directory as standalone class files.
But we can do that in a future cleanup PR as well.

flangofas: :+1:

flangofas:
@ADmad done.

ADmad:
Thanks @flangofas

ADmad:
@flangofas Would you mind updating the README too with examples showing how to use this new feature.

flangofas:
@ADmad :+1:

Issue #120, opened 2016-09-13
Better handling of pagination query string keeping


dereuromark:
Refs https://github.com/FriendsOfCake/search/pull/114

The docs example is wrong and overcomplicated

		foreach (\Cake\Utility\Hash::flatten($this->request->query) as $param => $value) {
			echo $this->Form->hidden($param, compact('value'));
		}

is missing the page omission needed, and it is quite overcomplicated.
The search plugin of cakedc handled this a bit better.
Can we maybe find a configure way to not needing the overhead here in the templates?

dereuromark:
Reclassifying as bug as the documented approach even breaks it all.
It currently would re-add the query string fields already present as POST which would always overwrite the other fields if coming after them (and the DOM is also invalid because the fields exist now twice).
We definitly need a better solution here.

inoas:
Opting out should be the way to persist even across multiple search filters + paginators per single request (Cells, etc):

```php
    foreach (\Cake\Utility\Hash::flatten($this->request->query) as $param => $value) {
        if (strpos($param, 'sort') !== false) { // This should respect the set scope, it's just meant as a POC
          echo $this->Form->hidden($param, compact('value'));
       }
    }
```

dereuromark:
Does not work, it will always reset to the previous ones, not the changed ones. And it is invalid DOM content. I mentioned all that above already.

PR #121, opened 2016-09-13
Bugfix keep sorting

dereuromark:
Resolves https://github.com/FriendsOfCake/search/issues/120

I wish there was a more automatic/smarter solution like the cakedc one had, but that would mean more of the logic needs to be moved to the component.
This is a minimal invasive approach to at least support it basically, and support it properly working.
Most importantly it does not include the page by default, which would only lead to big trouble.

### DIFF
--- a/src/Controller/Component/PrgComponent.php
+++ b/src/Controller/Component/PrgComponent.php
@@ -16,7 +16,7 @@ class PrgComponent extends Component
      *   as string or multiple as array. If boolean `true` all actions will be
      *   processed if `false` none. Default is ['index', 'lookup'].
      * - `queryStringToData` : Set query string as request data. Default `true`.
-     * - `queryStringWhitelist` : An array of whitelisted query strings to be kept.
+     * - `queryStringWhitelist` : An array of whitelisted query string keys to be kept.
      *   Defaults to the Paginator `'sort'` and `'direction'` ones.
      *
      * @var array
--- a/src/Controller/Component/PrgComponent.php
+++ b/src/Controller/Component/PrgComponent.php
@@ -2,6 +2,7 @@
 namespace Search\Controller\Component;
 
 use Cake\Controller\Component;
+use Cake\Controller\ComponentRegistry;
 use Cake\Utility\Hash;
 
 class PrgComponent extends Component
@@ -13,14 +14,15 @@ class PrgComponent extends Component
      * ### Options
      * - `actions` : Action name(s) to use PRG for. You can pass a single action
      *   as string or multiple as array. If boolean `true` all actions will be
-     *   processsed if `false` none. Default is ['index', 'lookup'].
+     *   processed if `false` none. Default is ['index', 'lookup'].
      * - `queryStringToData` : Set query string as request data. Default `true`.
      *
      * @var array
      */
     protected $_defaultConfig = [
         'actions' => ['index', 'lookup'],
-        'queryStringToData' => true
+        'queryStringToData' => true,
+        'queryStringWhitelist' => ['sort', 'direction']
     ];
 
     /**
@@ -31,9 +33,11 @@ class PrgComponent extends Component
      */
     public function startup()
     {
-        if ($this->_actionCheck()) {
-            return $this->conversion();
+        if (!$this->_actionCheck()) {
+            return null;
         }
+
+        return $this->conversion();
     }
 
     /**
@@ -57,7 +61,7 @@ class PrgComponent extends Component
 
         list($url) = explode('?', $this->request->here(false));
 
-        $params = Hash::filter($this->request->data);
+        $params = $this->_filterParams();
         if ($params) {
             $url .= '?' . http_build_query($params);
         }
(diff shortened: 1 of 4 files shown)


dereuromark:
I made a live example of that branch:
http://sandbox.dereuromark.de/sandbox/search-examples?search=a&status=1&page=8&sort=CountryRecords.iso3&direction=asc
And then make the "a" an "ab" search input content, which will bring you to
http://sandbox.dereuromark.de/sandbox/search-examples?search=ab&status=1&sort=CountryRecords.iso3&direction=asc
page is removed as expected.

ADmad:
LGTM but would be nice to have some tests for this so that there are not regressions in future.

dereuromark:
Tests added, also fixed up composer json dependencies.

src/Controller/Component/PrgComponent.php
```diff
@@ -13,14 +14,15 @@ class PrgComponent extends Component
...
+     *   processed if `false` none. Default is ['index', 'lookup'].
      * - `queryStringToData` : Set query string as request data. Default `true`.
      *
      * @var array
      */
     protected $_defaultConfig = [
         'actions' => ['index', 'lookup'],
-        'queryStringToData' => true
+        'queryStringToData' => true,
+        'queryStringWhitelist' => ['sort', 'direction']
```
inoas: `,` as far as the current outlook on that issue is.

src/Controller/Component/PrgComponent.php
```diff
@@ -31,9 +33,11 @@ class PrgComponent extends Component
      */
     public function startup()
     {
-        if ($this->_actionCheck()) {
-            return $this->conversion();
+        if (!$this->_actionCheck()) {
+            return null;
         }
+
+        return $this->conversion();
```
inoas: Does this work without the negative condition?
```php
public function startup()
{
    if ($this->_actionCheck()) {
        return $this->conversion();
    }

    return null;
}
```

inoas:
So as a whole this would persist sort and direction, but not page? What about any other params?
*Why not just remove the page and persist any other query string?*

src/Controller/Component/PrgComponent.php
```diff
@@ -13,14 +14,17 @@ class PrgComponent extends Component
      * ### Options
      * - `actions` : Action name(s) to use PRG for. You can pass a single action
      *   as string or multiple as array. If boolean `true` all actions will be
-     *   processsed if `false` none. Default is ['index', 'lookup'].
+     *   processed if `false` none. Default is ['index', 'lookup'].
      * - `queryStringToData` : Set query string as request data. Default `true`.
+     * - `queryStringWhitelist` : An array of whitelisted query strings to be kept.
```
inoas: query string keys

dereuromark:
Because that would destroy it again, as those can be from the previous search and would then force-keep every search value even if removed from the post.

src/Controller/Component/PrgComponent.php
dereuromark: Its called "return early".

src/Controller/Component/PrgComponent.php
dereuromark: ?

src/Controller/Component/PrgComponent.php
dereuromark: Too late.

src/Controller/Component/PrgComponent.php
inoas: Both are? Can we remove the negation in the if condition?

inoas:
@dereuromark If you paginate multiple tables on one single view (through cells maybe, and scopes) wouldn't those then get removed?
The snipplet I posted to the docs (that always persists page for any scope) just needs to opt-out the page param for any given scope (even no-scope):
See: https://github.com/FriendsOfCake/search/issues/120#issuecomment-247191204

dereuromark:
Your snippet does not work at all. For the reasons mentioned on the ticket.
And you can do that still if you want to, but a good, non-complicated and sane default solution here is the better way to go IMO.

inoas:
The solution should be opt-out not opt-in for persisting query strings.
You want to opt-out of persisting any page query string key-value pair as those can lead to 404 responses.

dereuromark:
So far the solution before worked as well, and now works even better with mine by persisting some whitelisted ones.
Opting out does not work here. Of course that would be better but cannot be done properly due to the reasons mentioned already.

dereuromark:
Waiting for merge and release so I can update my sandbox to use a tagged version here :)
http://sandbox.dereuromark.de/sandbox/search-examples
Feel free to add more advanced search examples to show the full power of the plugin :P

dereuromark:
Should limit param also be kept to be in sync with Paginator whilelist?

dereuromark:
Done, this should now be good to go.

ADmad:
@dereuromark This can be merged once docblock description is added for `_filterParams()`

dereuromark:
Pretty harsh there ;) Done.

ADmad:
I had already commented "Description please." above the method few days ago.

Issue #77, opened 2016-02-19
Multiple values


glafarge:
What do you think about the support of multiple values ?

For example, if i need the following configuration : 

```php
$this->addBehavior('Search.Search');
$this->searchManager()
    ->add('place', 'Search.Like', [
        'before'=>true,
        'after'=>true,
        'field' => [$this->aliasField('ZipCode'), $this->aliasField('City')]
]);
```

As the Like filter only supports single value, I have created a child filter to supports arrays.

```php
namespace App\Model\Filter;

use Cake\ORM\Query;

class LikeMultiple extends \Search\Model\Filter\Like
{

    /**
     * Process a LIKE condition ($x LIKE $y).
     * Allow multiple values
     *
     * @return void
     */
    public function process()
    {
        if ($this->skip()) {
            return;
        }

        $conditions = [];
        foreach ($this->fields() as $field) {
            $left = $field . ' ' . $this->config('comparison');

            if(is_array($this->value())) {
                foreach($this->value() as $value) {
                    $right = $this->_wildCards($value);
                    $conditions[] = [$left => $right];
                }
            }
            else {
                $right = $this->_wildCards($this->value());
                $conditions[] = [$left => $right];
            }
        }

        $this->query()->andWhere([$this->config('mode') => $conditions]);
    }
}
```

I think we can add this functionnality to Like filter directly. This can be extended to Compare too.
I can make a PR if needed.

What do you think about it and about security concerns ?

burzum:
> What do you think about the support of multiple values ?

Good idea.

But IMO it should be a method in the base class or a trait. I can see that you might want to use multiple values with an IN, but in another place you want to do an AND or an OR.

A param that can take N values should have a flag like `multiValue => true`.

glafarge:
What you suggest is to add functionality to Like base class, add a new boolean parameter to it named `multiValue`. So, we can deal with multiple values and avoid conflicts with IN filter when needed to. 
Right ?

burzum:
No.

1. The functionality should be optional
2. The feature should be easy to use by *any* filter
3. There should be some way to decide how to threat the multiple values (What you want to do with them)

So either make it a trait or the functionality can be implemented in https://github.com/FriendsOfCake/search/blob/master/src/Model/Filter/Base.php

glafarge:
Good ideas ! Anyway it needs more work to implement it.
Thank you, I'll look at it.

Mistermasque:
Hi,

[...]

PR #126, opened 2016-09-20
Add multi-value search support.

ndm2:
This is a first attempt of implementing #77

This change introduces the `modeValue` option in the `Base` filter. Support for it is being implemented in the `Value` and `Like` filters, it doesn't seem to make sense for any other filter, `Boolean` is binary, and for `Comparison` only one value only ever really would have an effect.

The `Base` filter exposes multiple values only in case the `multiValue` option is enabled, otherwise it will fall back to returning the default value, basically treating the arguments as non-existing.

For the `Value` filter, in `or` mode, `IN()` is kept being used, IIRC this is less parser/optimizer heavy than multiple `OR`s. For the `and` mode multiple `AND`s are being used, as for using `IN()` this would require grouping and using `HAVING`, and I'm not sure that such modifications are overly welcome.

The `Like` filter simply builds multiple (nested) `AND`/`OR`s.

Docs will be added once this is considered good to merge.

---

There a few things that not everyone may be sold on, like

* Deprecating the `mode` option (and triggering a notice :fearful:) for the `Like` filter, and introducing two separate options, `fieldMode` and `valueMode`, where the former is used for the possible multiple fields, and the latter for the possible multiple values.

  There are for sure other ways to handle this, like for example
  
  * Instead of deprecating, use the `mode` value as the default for `fieldMode` and `valueMode` in case they haven't explicitly been configured.
  
  * Again use `mode` as the default, but make it also accept an array with further separate keys to configure the mode for the fields and the values separately.

  * ...

* Testing the `findSearch` method in such an explicit manner.

* They way how the "_no default value_" situation is being handled, see the `testProcessNoDefaultFallbackForDisallowedMultiValue` tests.

### DIFF
--- a/src/Model/Filter/Like.php
+++ b/src/Model/Filter/Like.php
@@ -1,6 +1,8 @@
 <?php
 namespace Search\Model\Filter;
 
+use Search\Manager;
+
 class Like extends Base
 {
 
@@ -12,11 +14,34 @@ class Like extends Base
     protected $_defaultConfig = [
         'before' => false,
         'after' => false,
-        'mode' => 'or',
+        'mode' => null,
+        'fieldMode' => 'or',
+        'valueMode' => 'or',
         'comparison' => 'LIKE',
         'wildcardAny' => '*',
         'wildcardOne' => '?',
     ];
+
+    /**
+     * {@inheritDoc}
+     *
+     * @param string $name Name.
+     * @param \Search\Manager $manager Manager.
+     * @param array $config Config.
+     */
+    public function __construct($name, Manager $manager, array $config = [])
+    {
+        parent::__construct($name, $manager, $config);
+
+        $mode = $this->config('mode');
+        if ($mode !== null) {
+            trigger_error(
+                'The `mode` configuration option is deprecated, use `fieldMode` and `valueMode` instead.',
+                E_USER_DEPRECATED
+            );
+            $this->config('fieldMode', $mode);
+        }
+    }
 
     /**
      * Process a LIKE condition ($x LIKE $y).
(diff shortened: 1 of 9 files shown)


ADmad:
I am fine with deprecating `or` in `Like` but let's not trigger a notice. If set we can use it's value as value for `fieldMode`. Values as array didn't work earlier anyway so no need to use value of `mode` for `valueMode`.

dereuromark:
I would be fine with the exception as long as it is just a E_USER_DEPRECATED.
That way no harm, and there is only one way forward what to best use here (no confusion about multiple option keys etc).

src/Model/Filter/Like.php
```diff
@@ -29,31 +54,49 @@ public function process()
...
+            $left = $field . ' ' . $comparison;
+            if ($isMultiValue) {
+                $valueConditions = [];
+                foreach ($value as $val) {
+                    $right = $this->_wildcards($val);
+                    if ($right !== false) {
+                        $valueConditions[] = [$left => $right];
+                    }
+                }
+                if (!empty($valueConditions)) {
```
dereuromark: if ($valueConditions) {}
is usually a bit cleaner as it does not cloak the variable existence (when accidentally spelled wrong)

src/Model/Filter/Like.php
```diff
@@ -29,31 +54,49 @@ public function process()
...
+            } else {
+                $right = $this->_wildcards($value);
+                if ($right !== false) {
+                    $conditions[] = [$left => $right];
+                }
+            }
         }
 
-        $this->query()->andWhere([$this->config('mode') => $conditions]);
+        if (!empty($conditions)) {
```
dereuromark: Same here

src/Model/Filter/Like.php
inoas: `if ($valueConditions)` reads like a boolean. I pretty much prefer !empty($valueConditions)) and wished there was a "filled()" in native PHP.

dereuromark: "If" are bool decisions :) also see coding guidelines regarding less error prone code.

inoas: Sure but the check that is happening through php's messy types is actually !empty($valueConditions) when you do ($valueConditions) - right?

dereuromark: It is both totally the same thing, absolutely 0 difference from the result here, but one cloaks the warning, the other does not.

ndm2: I'd think that internally `($valueConditions)` is like check the type, compare size against 0, stirred with a shot of black PHP voodoo magic :sparkles: 

I'm fine with any of these variants. `count()` is like 2 to 4 times slower than the other too, in case that is of any concern.

ndm2:
So triggering a notice is +1/-1... any more votes or a final word? :)

ADmad:
@ndm2  Let's not trigger a notice, we have stopped doing so in core too.

ndm2:
Writing the docs for the options I noticed that the stuff around validation is never used, ie the `Base::validation()` and `Base::valid()` methods, and consequently the `valid` option.

Is this intentionally dead code that is ment to be removed at some point?

ndm2:
Ok, I'll leave out the `validate` option then.

dereuromark:
Maybe you can put `@deprecated` around those things instead, so it is clear that this will be removed with the next major.

ADmad:
Those methods should instead be marked `@internal` until the feature implementation is completed. But that's beyond the scope of this PR and can be dealt separately.

ndm2:
Another thing that I've noticed, the `wildcardAny` and `wildcardOne` options can be used to define custom strings to use as wildcards, which are then going to be replaced with `%` and `_` later on. If that happens, then possible existing `%` and `_` chars in the seach term are going to be escaped.

**https://github.com/FriendsOfCake/search/blob/2.2.0/src/Model/Filter/Like.php#L78**

So far so good, but there doesn't seem to be a way to disable accepting wildcards in the search term. Is this a lacking feature, or something that could be considered a bug? In case the default wildcards are not going to the escaped by default, this makes these options kinda superflous, doesn't it?

dereuromark:
If you use LIKE you would want to have wildcard capabilities, otherwise why not using value?
Do you have a test case that would display the issue? But this would also be out of scope of this PR probably.

ndm2:
Depending on the context one maybe want to restrict things to wildcard before/after only via options, ie don't let the user add wildcards as they please. Not sure how valid this use case is, it just popped to my mind when reading the code.

I'm just asking because I'm adding docs for all options now, not just the multi-value/mode related ones, I thought if I'm already at it, why not document all of them :)

dereuromark:
Well, I guess someone could enhance it to be false or null to be skipping the inline replacement part, but yeah, as you say, so far it does not seem to be too interesting to people.

ndm2:
Ok, I'll treat it as missing feature then.

ADmad:
@ndm2 The fact that you are reviewing the plugin code is greatly appreciated. We would love to have your opinions on it but please open separate tickets for any issues you find. That way the discussion around each issue can be easily tracked. Lets finish the job at hand and not turn it into this PR into a mega thread :)

ndm2:
I could hear myself saying that two comments ago :) I'll restrain myself now and add the docs around how things work currently.

ADmad:
No worries, keep at it :)

ADmad:
Thanks @ndm2

PR #127, opened 2016-09-22
Multi-value search support follow up

ndm2:
I missed some stuff in the docs, hardend the filters that do not support multi-values, as well as those that do.

One could argue that for example the `Value` filter should be able to receive non-flat multi-values, ie nested arrays, given that people could use custom database types that support such values and can convert them into something useable.

However, given that multi-values weren't supported before, and that such custom database types are most likely rather rare, I'd think that it's better to play safe, and have people that need more control create custom filters instead.


ADmad:
Why is the check to ensure all values in array are scalar necessary?

ndm2:
Generally this should ensure that in a multi-value situation, all passed values can actually be processed, so that for example queries with conditions like `column = ['value']` are being avoided.

Are you concernced about checking all values in general, or about checking for scalar instead of for example for non-array?

ndm2:
Having a further look at it, this check

**https://github.com/FriendsOfCake/search/blob/master/src/Model/Filter/Like.php#L94-L96**

could theoretically still cause partial conditions to be generated, in case the passed values do not directly stem from the query string, ie are string values. And given this string check, the multi-value validation is kinda useless here.

Say for example someone does something like

```php
$search = [
	'title' => [
		'foo',
		true
	]
];
$table->find('search', ['search' => $search]);
```

This would be considered a valid multi-value, but `true` would be discarded in the wildcard processor, causing only `foo` to make it into the conditions.

Not sure how valid this concern is, but I think that at least there should be protection against nested arrays, ie

```php
$search = [
	'title' => [
		'foo',
		['invalid']
	]
];
```

As this will most probably blow up things if it makes it into the type system for casting to a database value.

ADmad:
The valid multi value check looks like unnecessary overhead to me. If someone passes in invalid values you will end up with an invalid query and related exception. I really don't see a problem with that :)

ndm2:
Ok, let 'em crash hard :) Can I assume that you're also ok with the Like filter dropping only the invalid values instead of the whole search?

ADmad:
> Can I assume that you're also ok with the Like filter dropping only the invalid values instead of the whole search?

Yip

ndm2:
Gone diddly gone... I've kept an empty check for the Value filter though, as an empty list may be considered valid in terms of "do not search anything"!?

ADmad:
I realized one of my comments I had made in your previous PR was not address since you never saw it as I goofed up using the new review system :P  Would be better if we used capitalized `OR` instead of `or` for the `mode` option.

dereuromark:
Isn't everything lowercase so far, including 'or'? It would only be consistent to stay lowercased IMO.

ADmad:
Having sql keywords / fragments in caps provides better readability. For e.g. the `comparison` option has value `LIKE` not `like`. So internally all caps should be used. Users can still provides any casing they want and we convert them internally.

dereuromark:
I agree, internal handling is fine, just on the options it should be exposed as lowercase for consistency maybe.

ndm2:
I'm receiving mixed signals... @dereuromark did you mean to say _uppercase_? Or do you suggest to expose everything lowercase, ie also `LIKE` as `like`?

dereuromark:
I meant that internally it is fine to be UPPERcase, but the options in the configs to be set from the outside should be consistently lowercased, not some upper and some lower. Because from the outside it does not matter if those are ending up as SQL snippets or switch case keys etc.

ndm2:
Ah, ok, so in `$_defaultConfig` use uppercase, and in docs use lowercase?

dereuromark:
Even $_defaultConfig could be lowercase consistency like, as it will be uppercased anyway along the process and for internal SQL preparation.

src/Model/Filter/Boolean.php
```diff
@@ -25,7 +25,12 @@ public function process()
             return;
         }
 
-        $value = strtolower($this->value());
+        $value = $this->value();
+        if (!is_scalar($value)) {
+            return;
+        }
+
+        $value = strtolower($value);
```
ADmad: The default `falsy` lists contain boolean `false` too and `strtolower(false)` will return an empty string.

ADmad:
Core also uses uppercase in default arguments and it's uppercased internally too [ref](https://github.com/cakephp/cakephp/blob/master/src/Database/Expression/QueryExpression.php#L65).

So @ndm2 please change them to uppercase in code and documentation.

src/Model/Filter/Boolean.php
inoas: I think these options are good enough:
```

    protected $_defaultConfig = [
        'truthy' => [1, true, '1', 'true', 'yes', 'on'],
        'falsy' => [0, false, '0', 'false', 'no', 'off']
    ];
```
You can always extend those to also allow 'FALSE' or "OFF".

In case you really want to become case insensitive out of the box, this might work:
```php
if (is_string($value)) } {
  $value = strtolower($value);
}
```

dereuromark:
Maybe we should correct the core here, other places also use `$operators = ['and', 'or', 'xor'];` etc
And that is even very close to the DB layer.
Here it is still mainly input options that later "become" SQL operators, but until then those are just option values that do not need SHOUTING characters IMO.

src/Model/Filter/Boolean.php
ndm2: I'll blame it on whoever forgot to add tests for actual boolean values :stuck_out_tongue: I can get that fixed up too, the `is_string()` test should do it.

ADmad: > I'll blame it on whoever forgot to add tests for actual boolean values

That would be @inoas :stuck_out_tongue:

inoas: @ADmad @ndm2 I take the blame: https://github.com/FriendsOfCake/search/pull/128

ndm2: Ah, didn't saw that this was a new PR @inoas, I've pushed tests and a fix too :)

README.md
```diff
@@ -273,28 +273,28 @@ The following options are supported by all filters except `Callback` and `Finder
 - `after` (`bool`, defaults to `false`) Whether to automatically add a wildcard
   *after* the search term.
 
-- ~~`mode`~~ (`string`, default to `or`) **This options is deprecated**, please use
+- ~~`mode`~~ (`string`, default to `OR`) **This options is deprecated**, please use
```
dereuromark: defaults

src/Model/Filter/Boolean.php
inoas: Don't worry. Pick what you like. I (or @ADmad :p) will close whatever is left.

README.md
ADmad: I'll fix it post merge.

ndm2: @ADmad You can take the superfluous **s** from **options** :)

https://www.youtube.com/watch?v=KzyCi1BFATA

Issue #101, opened 2016-06-13
defaultConfig[mode] missing from Compare?


lamasgergo:
I think it's not a big deal to implement "mode" config into Compare. When you would like to compare multiple values with same field but with "OR" relation, now you can't.
For example you would like to search in a webshop for products with prices wich are stored in different coulumns.
Thanks! I love this plugin!

ADmad:
PR to implement the enhancement is welcome.

flangofas:
I can provide a fix :smile:

dereuromark:
PR is open.

PR #132, opened 2016-10-15
OR config param added to Compare

lamasgergo:
#101 improvement

### DIFF
--- a/src/Model/Filter/Compare.php
+++ b/src/Model/Filter/Compare.php
@@ -11,6 +11,7 @@ class Compare extends Base
      */
     protected $_defaultConfig = [
         'operator' => '>=',
+        'mode' => 'AND'
     ];
 
     /**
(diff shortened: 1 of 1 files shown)


dereuromark:
Could we please have a small test case for this addition?

burzum:
By looking at this, how about allowing to pass a callable as well and pass the query object on to it?

ADmad:
@burzum We already have a Callback filter, so if you need to use a callback just use that.

dereuromark:
@lamasgergo ping

lamasgergo:
i§m on it, sorry i don't have enough time for write it yet. It needed a simple test, when we have checking for example on the "created" AND "modified" fields. 

> select * from articles where created >= '2016-11-14' OR modified >= '2016-11-14' 

I try to publish this simple test ASAP.

dereuromark:
@lamasgergo Ok, thx!

lamasgergo:
Thanks @dereuromark !

PR #135, opened 2016-11-03
Allow multiple fields for value filter

steefaan:
I think it would be nice to provide the possibility to search a value inside multiple fields at once. This way the `Value` filter acts as `Like` filter. Code is BC, see the constructor.

### DIFF
--- a/src/Model/Filter/Value.php
+++ b/src/Model/Filter/Value.php
@@ -1,5 +1,4 @@
 <?php
-
 namespace Search\Model\Filter;
 
 use Cake\Database\Expression\QueryExpression;
@@ -7,6 +6,7 @@ use Search\Manager;
 
 class Value extends Base
 {
+
     /**
      * Default configuration.
      *
--- a/src/Model/Filter/Value.php
+++ b/src/Model/Filter/Value.php
@@ -1,17 +1,38 @@
 <?php
+
 namespace Search\Model\Filter;
+
+use Cake\Database\Expression\QueryExpression;
+use Search\Manager;
 
 class Value extends Base
 {
-
     /**
      * Default configuration.
      *
      * @var array
      */
     protected $_defaultConfig = [
-        'mode' => 'OR',
+        'fieldMode' => 'OR',
+        'valueMode' => 'OR'
     ];
+
+    /**
+     * {@inheritDoc}
+     *
+     * @param string $name Name.
+     * @param \Search\Manager $manager Manager.
+     * @param array $config Config.
+     */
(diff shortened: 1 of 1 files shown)


src/Model/Filter/Value.php
```diff
@@ -36,21 +57,25 @@ public function process()
...
-            foreach ((array)$value as $val) {
-                $e->eq($field, $val);
-            }
+                return $e;
+            };
+        }
 
-            return $e;
-        });
+        if (!empty($expressions)) {
```
dereuromark: I would leave out !empty()

dereuromark:
Good idea 👍

src/Model/Filter/Value.php
steefaan: It's a copy/paste from `Like` filter.

dereuromark: Yeah ideally they both remove the not necessary cloaking. Bot no big deal.

src/Model/Filter/Value.php
```diff
@@ -1,6 +1,9 @@
 <?php
 namespace Search\Model\Filter;
 
+use Cake\Database\Expression\QueryExpression;
+use Search\Manager;
+
 class Value extends Base
```
jippi: couldn't we make a MultiValue instead, this will otherwise be breaking change

steefaan: I would stay at `Value` otherwise we have an inconsistency between the usage of `Like` and `Value`. Where is the breaking change?

steefaan:
What about merging?

dereuromark:
As said before I am 👍 on merging this
Is anyone objecting to this?

ndm2:
Explicit multi field tests wouldn't hurt...

ADmad:
Tests and readme update required.

dereuromark:
@steefaan Are you able to finish it up or do you want some help here?

src/Model/Filter/Value.php
```diff
@@ -10,10 +13,28 @@ class Value extends Base
...
+     *
+     * @param string $name Name.
+     * @param \Search\Manager $manager Manager.
+     * @param array $config Config.
+     */
+    public function __construct($name, Manager $manager, array $config = [])
+    {
+        parent::__construct($name, $manager, $config);
+
+        $mode = $this->config('mode');
```
dereuromark: We could add an line comment here:

 > // For backwards compatibility

etc

dereuromark:
How can we finish this up? Anyone interested?

ndm2:
I'll finish it up...

PR #146, opened 2017-01-18
Resolves #145

### LINKED ISSUE #145: How to use a Finder type?
There seems to be no documentation on how to use a Finder type filter.

davidyell:


README.md
```diff
@@ -309,6 +309,10 @@ The following options are supported by all filters except `Callback` and `Finder
 - `mode` (`string`, defaults to `OR`) The conditional mode to use when searching for
   multiple values. Valid values are `OR` and `AND`.
 
+#### `Finder`
+
+- `finder` (`string`) The name of the finder method to use, without the 'find' prefix.
```
ADmad: "..without the 'find' prefix."  is not necessary. Instead link to the manual section about custom finders.

davidyell: So it's okay to use `'finder' => 'findByMonth'` ?

ADmad: You can't use `'finder' => 'findByMonth'`. Just link to docs, in any case if they know know how to write / use custom finder they will have to refer the cakephp docs, we are not going to duplicate them here.

davidyell: Telling someone how to call a method which your code executes, is not duplicating the documentation. It's documenting how your software works. Especially if there is possible ambiguity.

Issue #143, opened 2016-12-29
SQL Server LIKE filter escaping


stmeyer:
Hey,

when using a `LIKE` filter with sql server and the search string contains `_` the value is escaped `\_`. By searching for `example_text` it is searched for `example\_text`. But in SQL Server one has to specify the escape character with the `ESCAPE` keyword (https://msdn.microsoft.com/en-us/library/ms179859.aspx):
```sql
INSERT mytbl2 VALUES ('Discount is 10-15% off'), ('Discount is .10-.15 off');  
GO  
SELECT c1   
FROM mytbl2  
WHERE c1 LIKE '%10-15!% off%' ESCAPE '!';  
```
Do you know a way to send the `ESCAPE` keyword?

Thanks

stmeyer:
Any ideas? Maybe this function is not implemented in core. I can create issue in core if needed.

dereuromark:
I bet we can, based on the driver, use a specific behavior here.
Are you able to make a PR?

stmeyer:
closed because PR #149 is open

PR #149, opened 2017-02-14
SQL Server LIKE filter escaping #143

stmeyer:
Changes to search in SQL Server with better escaping. I check only the end of the class name, so that everone is able to derive a own implementation of the base class.


dereuromark:
Looks good to me.

ADmad:
I don't really like how the driver type guessing will be done on every call to `_formatWildcards()`. 

We can add a `driver` config to the filter. The driver guessing code can be moved into a separate method and called just from `process()` and set the `driver` config if not already set.

dereuromark:
Thats a valid concern. Sounds like a better approach.

src/Model/Filter/Like.php
```diff
@@ -1,12 +1,22 @@
...
+use InvalidArgumentException;
 use Search\Manager;
 
 class Like extends Base
 {
 
     /**
+     * driver to do escaping
+     *
+     * @var type
```
dereuromark: FQCN to type missing

src/Model/Filter/Like.php
```diff
@@ -11,7 +11,7 @@ class Like extends Base
     /**
      * driver to do escaping
      *
-     * @var type
+     * @var Search\Model\Filter\EscapeDriver\Base
```
dereuromark: FQCN starts with \

src/Model/Filter/EscapeDriver/Normal.php
```diff
@@ -0,0 +1,37 @@
+<?php
+namespace Search\Model\Filter\EscapeDriver;
+
+class Normal extends Base
```
dereuromark: Is "Default" a better name maybe?

dereuromark:
Personally, I dont like those short suffix-less class names (class Default extends Base).
I rather prefer sth like \Model\Filter\Escaper\ and
- BaseEscaper
- DefaultEscaper
- SqlserverEscaper

etc.
Those are more speaking and if search for easier to find between tons of "Default" or "Base" classes.

PS: I know that all the filters have the same issue and are missing the suffix. So for the sake of consistency I am fine with the current code - just saying I would have named it a bit differently.

stmeyer:
I added the suffix because I prefer it, too.

If this is ok now I will add some documentation for the escapers, because everyone is able to write on escapers if needed.

Additionally should I add a "NothingEscaper" that leaves the search string as is?

dereuromark:
Isnt nothing already possible by declaring the Wildcards to false? I would Not add it then twice.

src/Model/Filter/Like.php
```diff
@@ -1,12 +1,22 @@
...
 
 class Like extends Base
 {
 
     /**
+     * driver to do escaping
+     *
+     * @var \Search\Model\Filter\EscapeDriver\Base
+     */
+    protected $escapeDriver;
```
ADmad: Please change to `$_escapeDriver` for consistency.

stmeyer:
Any more hints or ready to merge?

ADmad:
There's lot of code duplication between `DefaultEscaper::formatWildCards()` and `SqlserverEscaper::formatWildCards()`. Would be nice if that could be improved.

Also I think we can do away with the `BaseEscaper` and just have `DefaultEscaper` and `SqlserverEscaper` with `SqlserverEscaper extends DefaultEscaper`.

src/Model/Filter/Like.php
```diff
@@ -143,26 +155,34 @@ protected function _wildcards($value)
...
         }
-        if ($from) {
-            // Escape first
-            $value = str_replace($from, $to, $value);
-            // Replace wildcards
-            $value = str_replace($substFrom, $substTo, $value);
+
+        $class = $this->config('escapeDriver');
+        $className = App::className($class, 'Model/Filter/EscapeDriver', 'Escaper');
+        if (!class_exists($className)) {
```
ADmad: `App::className` already does a `class_exists()` check and returns `false` on failure, so this should be just `if (!$className) {`

stmeyer:
last travis build did not update the status. are you able to update it?

src/Model/Filter/Like.php
```diff
@@ -1,12 +1,22 @@
...
+use InvalidArgumentException;
 use Search\Manager;
 
 class Like extends Base
 {
 
     /**
+     * driver to do escaping
+     *
+     * @var \Search\Model\Filter\EscapeDriver\Base
```
dereuromark: Base? Maybe we should have an interface then instead?

src/Model/Filter/Like.php
```diff
@@ -143,26 +155,34 @@ protected function _wildcards($value)
...
+    }
+
+    /**
+     * set configuration for escape driver name
+     *
+     * @return void
+     */
+    protected function _setEscapeDriver()
+    {
+        if ($this->config('escapeDriver') === null) {
```
dereuromark: Should this be "escaper" for consistency with the class name? Do we need the word "driver" in here? //cc @ADmad

dereuromark: Also the folder name doesnt have to be EscapeDriver but Escaper then if we change this.

src/Model/Filter/Like.php
dereuromark: formatWildcards() is the only method as per contract then.

stmeyer: Yes, interface is usefull

dereuromark:
Nice work!
Are you able to squash the commits?
I think we can otherwise also do this on merge with github.

stmeyer:
It would be nice if you can do it 😄

src/Model/Filter/Like.php
```diff
@@ -1,12 +1,22 @@
...
 
+use Cake\Core\App;
+use InvalidArgumentException;
 use Search\Manager;
 
 class Like extends Base
 {
 
     /**
+     * driver to do escaping
```
dereuromark: Escaper to be used.

src/Model/Filter/Like.php
```diff
@@ -51,6 +62,7 @@ public function process()
             return;
         }
 
+        $this->_setEscapeDriver();
```
dereuromark: Should this be $this->_setEscaper()?

dereuromark:
@ADmad This is good to go, right?

stmeyer:
@ADmad any news?

ADmad:
Sorry forgot about this. Will review again soon and merge.

src/Model/Filter/Like.php
```diff
@@ -1,12 +1,22 @@
...
 {
 
     /**
+     * Escaper to be used.
+     *
+     * @var \Search\Model\Filter\Escaper\EscaperInterface
+     */
+    protected $_escaper;
+
+
```
ADmad: Extra empty line

Issue #131, opened 2016-10-09
Search for ID


Isengo1989:
Sometimes it is faster to search for an ID. When I use it in the 'field' => array it won´t find the ID´s

dereuromark:
I don't understand. Do you have an example with some code?

Isengo1989:
Sorry didn´t notice until now. E-Mail came in the advertising folder.

**Controller**

```
public function index()
    {
        //$cat = $this->Auth->user('category');
        $this->paginate = ['order'=> ['Articles.display_date' => 'DESC']];

        $query = $this->Articles
                ->find('search', $this->Articles->filterParams($this->request->query));


            $this->set('articles', $this->paginate($query));
            return;
    
        }

```

**Model**

```
/**
     * Initialize method
     *
     * @param array $config The configuration for the Table.
     * @return void
     */
    public function initialize(array $config)
    {
        parent::initialize($config);

        // Add the behaviour to your table
        $this->addBehavior('Search.Search');

        // Setup search filter using search manager
        $this->searchManager()
            //->value('article_id')            
            ->add('q', 'Search.Like', [
                'before' => true,
                'after' => true,
                'mode' => 'or',
                'comparison' => 'LIKE',
                'wildcardAny' => '*',
                'wildcardOne' => '?',
                'field' => ['title','short_description','id']
            ])
            ->add('foo', 'Search.Callback', [
                'callback' => function ($query, $args, $filter) {
                    // Modify $query as required
                }
            ]);
```

**View**

```
<div class="search">
<?php
            echo $this->Form->create();
            echo '<span class="search-input">';
            echo $this->Form->input('q', ['label' => false, 'class'=>'form-control form-input','placeholder' => 'Suchbegriff']);
            echo '</span>';
            echo '<span class="search-button">';
            echo $this->Form->button('Suche', ['type' => 'submit', 'class' => 'btn btn-primary btn-m page-scroll']);
            //echo $this->Html->link('Reset', ['action' => 'index'], ['class' => 'buttonRight']);
            echo '</span>';
            echo $this->Form->end();
            ?>
</div>
```

ADmad:
Please provide the params you used and the query generated.

Isengo1989:
Not quiete sure what you mean by parameters, I guess how the request looks like:

https://www.test.de/subfolder/blog/alle-beitraege?q=Testartikel
**Works fine**

https://www.test.de/subfolder/blog/alle-beitraege?q=965342
[...]

PR #150, opened 2017-02-18
Allow numeric fields to be LIKE searched without auto-cast bug.

dereuromark:
Resolves the issue of silent auto-casting (and false positives) when doing string based LIKE searches.

Fixes https://github.com/FriendsOfCake/search/issues/131

Before with `->like('search', ['field' => ['name', 'iso2', 'iso3', 'country_code']]);`:
```sql
WHERE 
  (
    `Countries`.`name` like '123' 
    OR `Countries`.`iso2` like '123' 
    OR `Countries`.`iso3` like '123' 
    OR `Countries`.`country_code` = 0
  ) 
```
After:
```sql
WHERE 
  (
    `Countries`.`name` like '123' 
    OR `Countries`.`iso2` like '123' 
    OR `Countries`.`iso3` like '123' 
    OR `Countries`.`country_code` like '123'
  ) 
```

### DIFF
--- a/src/Model/Filter/Like.php
+++ b/src/Model/Filter/Like.php
@@ -56,7 +56,12 @@ class Like extends Base
         $isMultiValue = is_array($value);
 
         $conditions = [];
+        $types = [];
         foreach ($this->fields() as $field) {
+            if ($this->_requiresTypeMapping($field)) {
+                $types[$field] = 'string';
+            }
+
             $left = $field . ' ' . $comparison;
             if ($isMultiValue) {
                 $valueConditions = [];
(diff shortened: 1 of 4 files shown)


dereuromark:
The fails are due to phpunit6 etc it seems, unrelated to the changes.

ADmad:
I don't like the overhead of checking column type for each field. A much simpler solution to the problem would be to add new config `colType` to override the column type.

```php
->like(
    'search', 
    [
        'field' => ['name', 'iso2', 'iso3', 'country_code'], 
        'colType' => ['country_code' => 'string']
    ]);
```

Then the only change you need in filter code is
```php
$this->query()->andWhere([$this->config('fieldMode') => $conditions], $this->config('colType') ?: []);
```

dereuromark:
I like the default fixing here, especially as this is a hidden issue in most cases where the autocast happens silently.
Maybe we could add both?
By default it checks the config, and if nothing is found (and not bool false) it does the auto lookup.
I think that is a sane default. Plus the overhead here is irrevelant as it only happens once per request in a crud based environment, so forms and alike build here too and are 100x slower overall.

dereuromark:
I modified and also added some docs for README.

src/Model/Filter/Like.php
```diff
@@ -78,11 +85,33 @@ public function process()
...
     /**
+     * @param string $field Field
+     * @return bool
+     */
+    protected function _requiresTypeMapping($field)
+    {
+        if (strpos($field, '.') === false) {
+            return false;
+        }
+        list($model, $field) = explode('.', $field);
```
dereuromark: This could still find the wrong model here, right? I dont quite like it yet.
The actual base class should be doing this prior to aliasing the field probably.
Any ideas here?

ADmad:
Closing as #152 has been merged.

inoas:
This is really good. So far I have always used value comparison for integers because of this. Thank you so much for the PR @dereuromark

PR #152, opened 2017-02-18
Manual version: Allow numeric fields to be LIKE searched

dereuromark:
Alternative to https://github.com/FriendsOfCake/search/pull/150 
I dont like it that this has to be done manually, but so far the autodetecting is not bullet proof and therefore this is probably the safer solution for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,7 +271,8 @@ The following options are supported by all filters except `Callback` and `Finder
   using the conditional operator defined by the `fieldMode` option.
   
 - `colType` (`array`), Use to set a custom type for any column that needs to be treated as
-  string column despite its actual type.
+  string column despite its actual type. This is important for integer fields, for example, if they
+  are part of the fields to be searched.
 
 - `before` (`bool`, defaults to `false`) Whether to automatically add a wildcard
   *before* the search term.
--- a/README.md
+++ b/README.md
@@ -264,11 +264,14 @@ The following options are supported by all filters except `Callback` and `Finder
   multiple values. If disabled, and multiple values are being passed, the filter
   will fall back to using the default value defined by the `defaultValue` option.
 
-- `field` (`string|array`, defaults to the name passed to the first argument of the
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -264,11 +264,14 @@ The following options are supported by all filters except `Callback` and `Finder
...
 
-- `field` (`string|array`, defaults to the name passed to the first argument of the
+- `field` (`string|array`), defaults to the name passed to the first argument of the
   add filter method) The name of the field to use for searching. Works like the base
   `field` option but also accepts multiple field names as an array. When defining
   multiple fields, the search term is going to be looked up in all the given fields,
   using the conditional operator defined by the `fieldMode` option.
+  
+- `colType` (`array`), Use to set a custom type for any column that needs to be treated as
+  string column despite its actual type.
```
ADmad: The description could be expanded to mention that you should use this feature if you want to do "Like" searches on integer columns.

src/Model/Filter/Like.php
```diff
@@ -56,6 +57,7 @@ public function process()
         $isMultiValue = is_array($value);
 
         $conditions = [];
+        $colTypes = $this->config('colType') ?: [];
```
ADmad: The `?: []` is not necessary as the default value for `colType` is already set to `[]`. Also the value can be fetched inside `if (!empty($conditions)) {` itself.

src/Model/Filter/Like.php
```diff
@@ -78,11 +80,34 @@ public function process()
...
+            if ($colTypes) {
+                $colTypes = $this->_aliasColTypes($colTypes);
+            }
+
+            $this->query()->andWhere([$this->config('fieldMode') => $conditions], $colTypes);
         }
     }
 
     /**
+     * @param array $colTypes Column types
```
ADmad: Description please.

ADmad:
Apart from the few minor things I pointed out this looks good :+1:

dereuromark:
All done.

PR #154, opened 2017-02-27
Fix up composer description

dereuromark:
Don't use the misleading word framework for a plugin, and also dont use the exact version as this lies now.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
     "name": "friendsofcake/search",
-    "description": "CakePHP 3.0 easy search framework",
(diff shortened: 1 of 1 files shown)


jippi:
don't see why framework is misleading for the plugin, it provides infrastructure for a specific type of application system, allowing you to easily extend it.. CRUD would be a framework too :P

dereuromark:
The version number is already provided by the rest of the meta information. But the plugin can sure be used with other minors and soon also majors. No need to have a description that is soon to be wrong IMO.

dereuromark:
Good to Go then?

PR #156, opened 2017-03-05
Use query interface to allow for other data sources

Marlinc:
This would technically speaking break BC on the `query` method.

### DIFF
--- a/src/Model/Filter/Base.php
+++ b/src/Model/Filter/Base.php
@@ -2,7 +2,7 @@
 namespace Search\Model\Filter;
 
 use Cake\Core\InstanceConfigTrait;
-use Cake\ORM\Query;
+use Cake\Datasource\QueryInterface;
 use Search\Manager;
 
 /**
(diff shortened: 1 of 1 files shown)


dereuromark:
We should First Release a 2.4 and then we can merge and Release 3.0 here.

dereuromark:
What is the plan of attack here?

Marlinc:
I think its best to create a new release as this is a breaking change. I'm not sure though how many people actually override the query method.

Do you see any other things that I have to change before this is mergeable?

jippi:
if we merge it we should just major bump, versions are cheap :)

Marlinc:
Should I target another branch?

dereuromark:
No, all good. We merge prior to the bump.

Marlinc:
Okay, anything else I should do?

ADmad:
I can merge this and bump version after other actionable open PRs are taken care of.

PR #159, opened 2017-04-06
Change to accessor methods

davidyell:
Updated the documentation to reflect the accessor method changes to the request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ public function index()
     $query = $this->Articles
         // Use the plugins 'search' custom finder and pass in the
         // processed query params
-        ->find('search', ['search' => $this->request->query])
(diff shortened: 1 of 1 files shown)


dereuromark:
I dont think the people using anything below 3.4 (3.0-3.3) would appreciate this change.
Plugins should be careful on how to limit the user base and the supporting core versions.
This change can be done on project level - but here it would bring no value but cut off some of the not upgraded people of getting updates 

:-1: until we must 3.4+ for some functional reason.

ADmad:
Right, this won't do any good and just increases the potential of "readme instructions don't work" type of tickets.

davidyell:
Yep, I'll take those point. Fair enough. 😄

PR #160, opened 2017-04-21
Allow emptyFields for PRG component in order for isSearch to stay correct

dereuromark:
This came up a few times before as an issue if I recall correctly.

In my case
```php
	echo $this->Form->input('search');
	echo $this->Form->input('has_missing_translation', ['type' => 'checkbox']);
```

The 0 posted will always result in the URL `?has_missing_translation=0`, which then also affects the isSearch and the reset button in the process.
The empty values should be seen as such (as the CakeDC one did). This is a non-invasive minimal BC safe addition that resolves the issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -375,6 +375,20 @@ Be sure to allow empty in your search form, if you're using one.
 echo $this->Form->input('author_id', ['empty' => 'Pick an author']);
 ```
 
+## Empty fields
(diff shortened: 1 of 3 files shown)


dereuromark:
Is this OK as minor prior to the major bump? Or include it in the other two PRs?

dereuromark:
Thank you!

PR #161, opened 2017-04-25
Add multi-field search support to value filter.

ndm2:
Basically a copy of #135, but complete with tests.

However, I've just noticed that having a mode for the values doesn't seem to make any sense, given that this filter applies an EQ comparison, applying AND to multiple values would never match, as a field can only hold one of the values in such a comparison.

Am I missing something, or did I just made a mistake when I added this initially, and nobody noticed it?

### DIFF
--- a/tests/TestCase/Model/Filter/ValueTest.php
+++ b/tests/TestCase/Model/Filter/ValueTest.php
@@ -22,6 +22,20 @@ class ValueTest extends TestCase
     /**
      * @return void
      */
+    public function testDeprecatedModeOption()
+    {
+        $articles = TableRegistry::get('Articles');
+        $manager = new Manager($articles);
+        $filter = new Value('title', $manager, ['mode' => 'modeValue']);
+
+        $this->assertEquals('modeValue', $filter->config('mode'));
+        $this->assertEquals('modeValue', $filter->config('valueMode'));
+        $this->assertEquals('OR', $filter->config('fieldMode'));
+    }
+
+    /**
+     * @return void
+     */
     public function testSkipProcess()
     {
         $articles = TableRegistry::get('Articles');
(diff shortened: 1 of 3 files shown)


ADmad:
> However, I've just noticed that having a mode for the values doesn't seem to make any sense, given that this filter applies an EQ comparison, applying AND to multiple values would never match, as a field can only hold one of the values in such a comparison.

You are right, using `AND` for multiple values make sense, it would never match. Seems we did over look that.

dereuromark:
Yes, it only makes sense for the field part.

ndm2:
Given that it would be kind of a bug fix, would you be OK with a "hard break" in terms of removing that feature/option, ie only use `IN()` for multiple values?

dereuromark:
Lets skip it together with https://github.com/FriendsOfCake/search/pull/156 as major bump. All good then.

ADmad:
I don't think we are breaking anything as `'mode' => 'AND' `  would have never "worked" in the first place.

ndm2:
That would be my stance on it too, just wanted to ask as people have different opinions when it comes to removing not-correctly/non working, but possibly in use behavior. I'll remove it.

tests/TestCase/Model/Filter/ValueTest.php
```diff
@@ -215,13 +317,13 @@ public function testProcessNoDefaultFallbackForDisallowedMultiValue()
...
      */
-    public function testProcessCaseInsensitiveMode()
+    public function testProcessCaseInsensitiveValueMode()
     {
         $articles = TableRegistry::get('Articles');
         $manager = new Manager($articles);
         $filter = new Value('title', $manager, [
             'multiValue' => true,
-            'mode' => 'Or'
+            'valueMode' => 'Or'
```
dereuromark: above you used "mode".

dereuromark:
@ADmad Do you want to tag a new minor before we merge and release the new major change?

PR #162, opened 2017-04-27
Complete multi field search support

ndm2:
Adds multi field support to the boolean filter too, now things should be complete.

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,6 +253,11 @@ The following options are supported by all filters except `Callback` and `Finder
 - `defaultValue` (`mixed`, defaults to `null`) The default value that is being
   used in case the value passed for the corresponding field is invalid or missing.
 
+### `Boolean`
+
+- `mode` (`string`, defaults to `OR`) The conditional mode to use when matching
+  against multiple fields. Valid values are `OR` and `AND`.
+
 #### `Compare`
 
 - `operator` (`string`, defaults to `>=`) The operator to use for comparison. Valid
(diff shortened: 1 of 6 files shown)


tests/TestCase/Model/Filter/ValueTest.php
```diff
@@ -46,6 +46,25 @@ public function testSkipProcess()
     /**
      * @return void
      */
+    public function testProcessWithoutField()
+    {
+        $articles = TableRegistry::get('Articles');
+        $manager = new Manager($articles);
+        $filter = new Value('title', $manager, [
+            'field' => []
```
ADmad: Shouldn't a filter configured without any field be treated as dev error and an exception be thrown?  I don't see a use case where empty field list would make sense.

ndm2: That's a good point. Initially I've added it for the other filters in the first round of the multi-field support changes, in order to avoid invalid queries being generated, but since I forgot to add specific tests for that behavior, I now don't know the scenario anymore in which I saw this happening :(

I'd agree that throwing an exception would probably the appropriate thing to do, however that would be a BC break at this point I guess?

ndm2: After a second look I think I now remember, the like filter needed some kind of protection because of possibly invalid/non-string values, and this protection could cause an empty set of conditions, ie this was actually a different situation.

Then the value filter got it too, as it was copied from the like filter, then I copied the value filter, from which in turn I spread it anywhere, and here we are now :)

So I guess in order to not break with the current behavior, that is, invalid queries being generated, causing a failure, I should revert this for all but the like filter, and then maybe an exception could be added for the next major?

ADmad: We just bumped the major version to 3.0.0. I think it's fine to add the exceptions for cases where it's surely dev error and have them included of 3.0.1.

ADmad:
@ndm2 Test failing.

PR #164, opened 2017-04-28
Api improvement for isSearch.

dereuromark:
After be just bumped the major, I think this should also land in that new version.
It is BC - and improves the API by clearly returning a bool result on process() to indicated if a search filter has been actually applied.
It also simplifies the code since skip() doesnt need to be invoked twice.

### DIFF
--- a/src/Model/Behavior/SearchBehavior.php
+++ b/src/Model/Behavior/SearchBehavior.php
@@ -214,10 +214,13 @@ class SearchBehavior extends Behavior
             $filter->setArgs($params);
             $filter->setQuery($query);
 
-            if (!$filter->skip()) {
+            if ($filter->skip()) {
+                continue;
+            }
+            $result = $filter->process();
+            if ($result !== false) {
                 $this->_isSearch = true;
             }
-            $filter->process();
         }
 
         return $query;
--- a/src/Model/Filter/Value.php
+++ b/src/Model/Filter/Value.php
@@ -18,24 +18,20 @@ class Value extends Base
     /**
      * Process a value condition ($x == $y).
      *
-     * @return void
+     * @return bool
      */
     public function process()
     {
-        if ($this->skip()) {
-            return;
-        }
-
         $value = $this->value();
         if ($value === null) {
-            return;
+            return false;
         }
 
         $isMultiValue = is_array($value);
         if ($isMultiValue &&
             empty($value)
         ) {
-            return;
+            return false;
         }
 
         $expressions = [];
@@ -49,6 +45,10 @@ class Value extends Base
             };
         }
 
-        $this->getQuery()->andWhere([$this->config('mode') => $expressions]);
+        if (!empty($expressions)) {
+            $this->getQuery()->andWhere([$this->config('mode') => $expressions]);
+        }
+
+        return true;
     }
 }
--- a/src/Model/Filter/Value.php
+++ b/src/Model/Filter/Value.php
@@ -45,9 +45,7 @@ class Value extends Base
             };
         }
 
-        if (!empty($expressions)) {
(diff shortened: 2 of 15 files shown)


src/Model/Behavior/SearchBehavior.php
```diff
@@ -214,10 +214,13 @@ protected function _processFilters($filters, $params, $query)
             $filter->args($params);
             $filter->query($query);
 
-            if (!$filter->skip()) {
+            if ($filter->skip()) {
+                continue;
+            }
+            $result = $filter->process();
+            if ($result !== false) {
```
ADmad: Explicit check for `false` will cause problems with existing custom filters people have which don't return anything.

dereuromark: I did this on purpose. All current would Return null and behave Like true.

dereuromark: This is already the current behavior. So it only Got better now, Not worse.

ADmad: Ah right it checks for not false. This is why i dislike negations in general, easy to misread :) Valid use here.

dereuromark: :) All good.

dereuromark:
I rebased and resolved the conflict.

dereuromark:
I suggest the following release notes for the tag 3.0.1:

```
### Improvements

Filters
- All filters should now return a boolean result on whether they modified the query. This allows for a more correct isSearch value.

**Comment by josegonzalez** [comment]
Why all the test removals?

**Comment by dereuromark** [comment]
Because they are obsolete with this improvement. You don't have to check for the skip being called.

**Comment by dereuromark** [comment]
I rebased it - should be clear now.

**Comment by josegonzalez** [comment]
Not bc as it changes the signature of a method. I think this is okay to go into 3.1.0 though.

**Comment by dereuromark** [comment]
The idea was to release a 3.0.1 for it all after the release of 3.0.0 was just the other day.

**Comment by dereuromark** [comment]
But both is fine.

**Comment by dereuromark** [comment]
@josegonzalez Strictly speaking it is BC, the method signature only changes on the semantic level.
All custom filters would behave the same old way - false positive report.
So it enhanced it by adding a more accurate report - that's all. Not a BC break as I see it.

File: src/Model/Filter/Value.php [hunk 2]
```diff
@@ -45,9 +45,7 @@ public function process()
             };
         }
 
-        if (!empty($expressions)) {
-            $this->getQuery()->andWhere([$this->config('mode') => $expressions]);
-        }
+        $this->getQuery()->andWhere([$this->config('mode') => $expressions]);
```
dereuromark: Did I rebase wrongly?

josegonzalez: We removed the empty check in favor of validation expressions.

josegonzalez: https://github.com/FriendsOfCake/search/commit/5bb12bd9bcfb4ec4ac59a5e13cf3a96532f7f6c9

josegonzalez:
Strictly speaking, `void` is not `bool`.

src/Model/Filter/Value.php
dereuromark: Ah thx.

dereuromark:
We both know what strictly means in a PHP world pre-7.1 - where we do not typehint anything scalar right now.
We do not need to fight about semantic meanings here - as it bears no relevance on the functional BC I was talking about ;)

dereuromark:
And just to be clear: A change to bool is always BC from void - not the other way around.
So you just made the point for me: enhancement, not a BC break.

josegonzalez:
Sorry, I think we'll agree to disagree.

In any case, this is good to go.

PR #165, opened 2017-04-28
Replace combined getter/setter

dereuromark:
for a cleaner API.
BC due to kept old methods - just added deprecation doc blocks.

Also fixed valid() to return a proper bool result.


josegonzalez:
Needs rebase, can you take care of that?

PR #166, opened 2017-04-28
Split combined setters/getters for cleaner API.

dereuromark:
Continue https://github.com/FriendsOfCake/search/pull/165

The different return types of useCollection() vs getCollection is especially important to clean up here.

### DIFF
--- a/src/Manager.php
+++ b/src/Manager.php
@@ -127,10 +127,10 @@ class Manager
     public function collection($name = null)
     {
         if ($name === null) {
-            return $this->useCollection();
+            return $this->getCollection();
         }
 
-        return $this->setCollection($name);
+        return $this->useCollection($name);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Manager.php
```diff
@@ -65,12 +65,12 @@ public function repository()
     /**
      * Return Table
      *
-     * @return \Cake\ORM\Table Table Instance
+     * @return \Cake\Datasource\RepositoryInterface Repository Instance
```
josegonzalez: This does change the return and it's not BC per-se, but I'm going to merge this in as it's the proper way to do things.

dereuromark: It fixes only the doc block (semantic) to what the non-deprecated method already returns.

josegonzalez: Yep thats why I'm fine with merging as is.

PR #171, opened 2017-07-14
Allow passing un-flattened params to filters (closes #170)

### LINKED ISSUE #170: Callback & Finder don't handle value passed if it is a keyed array
I'll start with an example of what works and what doesn't because it will make what I'm trying to do a little clearer:

This example shows the current behavior which works:
```php

//request data contains an array like this from a multi-select field
$requestData = [
    'sp_id' => [
        0 => '123',
        1 => '456'
    ]
];

//filter setup
$this->searchManager()->add('sp_id', 'Search.Callback', [
    'callback' => function ($query, $args, $filter) {
        //execution get's here, all is well, I can process the array in $args
    }
]);
```

The following DOESN'T work, but it's what I'm trying to do
```php

//request data contains a keyed array from a date-select field
$requestData = [
    'start_date' => [
        'year' => '2014',
        'month' => '01',
        'day' => '13'
    ]
];

//filter setup
$this->searchManager()->add('start_date', 'Search.Callback', [
    'callback' => function ($query, $args, $filter) {
        //execution NEVER gets here, the array is totally ignored!
    }
]);
```

In the above examples I would expect each scenario to behave the same way; to pass along the array to my callback or finder and let me deal with parsing the data. However the second situation doesn't work at all, my callback or finder are never called if the value is a keyed array.

I'm trying to make a filter to narrow some rows down by date, and the second example shows how my form gives the data.

JayWalker512:
This allows filters to deal with dates in the default way the CakePHP FormHelper delivers them, ie:
```php
$dateInRequest = [
    'someDate' => [
        'year' => '2014',
        'month' => '01',
        'day' => '13'
    ]
];
```
which lets me configure my filter like so:
```php
$this->searchManager() ->add('start_date', 'Search.Callback', [
    'flatten' => false,
    'callback' => function ($query, $args, $filter) {
        $arrayKeysExist = function (array $keys, array $arr) {
            return !array_diff_key(array_flip($keys), $arr);
        };

        $parts = $args[$filter->getConfig()['name']];
        if ($arrayKeysExist(['year', 'month', 'day'], $parts)) {
            $startDate = new Time(date('Y-m-d', strtotime('now')));
            $timestamp = $startDate->year($parts['year'])->month($parts['month'])->day($parts['day'])->toUnixString();
            $query->where([
                'UNIX_TIMESTAMP(Timesheets.start_time) >' => $timestamp
            ]);
        }
        return $query;
    }
```
Likewise I can go on to write a class for filtering dates like this, but this is a good proof of concept.

### DIFF
--- a/src/Model/Behavior/SearchBehavior.php
+++ b/src/Model/Behavior/SearchBehavior.php
@@ -166,7 +166,7 @@ class SearchBehavior extends Behavior
     {
         $flattened = [];
         foreach ($params as $key => $value) {
-            if (is_array($value) && !empty($filters[$key]) && $filters[$key]->getConfig()['flatten'] == false) {
+            if (is_array($value) && !empty($filters[$key]) && $filters[$key]->getConfig()['flatten'] === false) {
                 $flattened[$key] = $value;
                 continue;
             }
--- a/src/Model/Behavior/SearchBehavior.php
+++ b/src/Model/Behavior/SearchBehavior.php
@@ -166,21 +166,23 @@ class SearchBehavior extends Behavior
     {  
         $flattened = [];
         foreach ($params as $key => $value) {
+            if (is_array($value) && !empty($filters[$key]) && $filters[$key]->getConfig()['flatten'] == false) {
+                $flattened[$key] = $value;
+                continue;
+            }
+            
             if (is_array($value)) {
-                if (!empty($filters[$key]) && $filters[$key]->getConfig()['flatten'] == false) {
-                    $flattened[$key] = $value;
-                } else {
-                    foreach ($value as $childKey => $childValue) {
-                        if (!is_numeric($childKey)) {
-                            $flattened[$key . '.' . $childKey] = $childValue;
-                        } else {
-                            $flattened[$key][$childKey] = $childValue;
-                        }
+                foreach ($value as $childKey => $childValue) {
+                    if (!is_numeric($childKey)) {
+                        $flattened[$key . '.' . $childKey] = $childValue;
+                    } else {
+                        $flattened[$key][$childKey] = $childValue;
                     }
                 }
-            } else {
-                $flattened[$key] = $value;
+                continue;
             }
+            
+            $flattened[$key] = $value;
         }
 
         return $flattened;
--- a/src/Model/Behavior/SearchBehavior.php
+++ b/src/Model/Behavior/SearchBehavior.php
@@ -62,9 +62,9 @@ class SearchBehavior extends Behavior
                 'to be nested under key "search" in find() options.'
             );
         }
-        
+
         $filters = $this->_getAllFilters(Hash::get($options, 'collection', 'default'));
-  
+
         $params = $this->_flattenParams((array)$options['search'], $filters);
         $params = $this->_extractParams($params, $filters);
 
@@ -163,26 +163,26 @@ class SearchBehavior extends Behavior
      * @return array The flattened parameters array.
      */
     protected function _flattenParams($params, $filters)
-    {  
+    {
         $flattened = [];
         foreach ($params as $key => $value) {
             if (is_array($value) && !empty($filters[$key]) && $filters[$key]->getConfig()['flatten'] == false) {
                 $flattened[$key] = $value;
                 continue;
             }
-            
-            if (is_array($value)) {
-                foreach ($value as $childKey => $childValue) {
-                    if (!is_numeric($childKey)) {
-                        $flattened[$key . '.' . $childKey] = $childValue;
-                    } else {
-                        $flattened[$key][$childKey] = $childValue;
-                    }
-                }
+
+            if (!is_array($value)) {
+                $flattened[$key] = $value;
                 continue;
             }
-            
-            $flattened[$key] = $value;
+
+            foreach ($value as $childKey => $childValue) {
+                if (!is_numeric($childKey)) {
+                    $flattened[$key . '.' . $childKey] = $childValue;
+                } else {
+                    $flattened[$key][$childKey] = $childValue;
+                }
+            }
         }
 
         return $flattened;
--- a/src/Model/Behavior/SearchBehavior.php
+++ b/src/Model/Behavior/SearchBehavior.php
@@ -166,15 +166,9 @@ class SearchBehavior extends Behavior
     {
         $flattened = [];
         foreach ($params as $key => $value) {
-            if (is_array($value)
-                && !empty($filters[$key])
-                && $filters[$key]->config()['flatten'] === false
+            if (!is_array($value) ||
+                (!empty($filters[$key]) && $filters[$key]->config()['flatten'] === false)
             ) {
-                $flattened[$key] = $value;
-                continue;
-            }
-
-            if (!is_array($value)) {
                 $flattened[$key] = $value;
                 continue;
             }
--- a/src/Model/Behavior/SearchBehavior.php
+++ b/src/Model/Behavior/SearchBehavior.php
@@ -65,13 +65,11 @@ class SearchBehavior extends Behavior
         
         $filters = $this->_getAllFilters(Hash::get($options, 'collection', 'default'));
 
-        $flattenWhitelist = []; //TODO FIXME should this actually be an array?
+        $flattenWhitelist = [];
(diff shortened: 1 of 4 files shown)


src/Model/Behavior/SearchBehavior.php
```diff
@@ -62,10 +62,18 @@ public function findSearch(Query $query, array $options)
                 'to be nested under key "search" in find() options.'
             );
         }
-
+        
         $filters = $this->_getAllFilters(Hash::get($options, 'collection', 'default'));
 
-        $params = $this->_flattenParams((array)$options['search']);
+        $flattenWhitelist = [];
+        foreach($filters as $filter) {
```
ADmad: This should be done in `_flattenParams()` method itself. Pass `$filters` as second argument to it.

ADmad: Also the filters are keyed by filter name in `$fitlers` so you can avoid this extra loop all together. You can just use `$filters[$key]` (with not empty check) to get filter config inside the loop used in `_filterParams()`.

JayWalker512: Ah good points, thanks. I will take care of this in a week as I'm traveling at the moment.

JayWalker512:
These last commits hit the points you mentioned. If you're happy with this, I'll update the readme as well to document the new option before merging.

src/Model/Behavior/SearchBehavior.php
```diff
@@ -167,19 +159,23 @@ protected function _extractParams($params, $filters)
...
-            if (is_array($value) && !in_array($key, $flattenWhitelist)) {
-                foreach ($value as $childKey => $childValue) {
-                    if (!is_numeric($childKey)) {
-                        $flattened[$key . '.' . $childKey] = $childValue;
-                    } else {
-                        $flattened[$key][$childKey] = $childValue;
+            if (is_array($value)) {
+                if (!empty($filters[$key]) && $filters[$key]->getConfig()['flatten'] == false) {
+                    $flattened[$key] = $value;
+                } else {
```
dereuromark: continue;
to avoid deeper nesting

JayWalker512: Is this the sort of less-nested structure you're suggesting?

```php
protected function _flattenParams($params, $filters)
{  
    $flattened = [];
    foreach ($params as $key => $value) {
        if (is_array($value) && !empty($filters[$key]) && $filters[$key]->getConfig()['flatten'] == false) {
            $flattened[$key] = $value;
            continue;
        }
            
        if (is_array($value)) {
            foreach ($value as $childKey => $childValue) {
                if (!is_numeric($childKey)) {
                    $flattened[$key . '.' . $childKey] = $childValue;
                } else {
                    $flattened[$key][$childKey] = $childValue;
                }
            }
            continue;
        }
            
        $flattened[$key] = $value;
    }

    return $flattened;
}
```

dereuromark: Yeah
you could even invert the is_array() part of the 2nd deeply nested one
Alternatively, speaking protected submethods can also take out some of the complexity.

JayWalker512: The second `is_array()` check is necessary to ensure the original behavior is maintained, where an array passed with no filter configuration is flattened as a matter of course. Here is how the original method looked before I changed anything:
```php
protected function _flattenParams($params)
{
    $flattened = [];
    foreach ($params as $key => $value) {
        if (is_array($value)) {
            foreach ($value as $childKey => $childValue) {
                if (!is_numeric($childKey)) {
                    $flattened[$key . '.' . $childKey] = $childValue;
                } else {
                    $flattened[$key][$childKey] = $childValue;
                }
            }
        } else {
            $flattened[$key] = $value;
        }
    }
    return $flattened;
}
```
 I think the complexity has only increased minimally, and is no more nested than before.

dereuromark: because you dont bail early where you can :)

e.g. if (is_array($value)) {} could be

    if (!is_array($value)) {
        continue;
    }

    foreach ....

src/Model/Behavior/SearchBehavior.php
```diff
@@ -62,10 +62,10 @@ public function findSearch(Query $query, array $options)
                 'to be nested under key "search" in find() options.'
             );
         }
-
+
```
dereuromark: You are introducing noise whitespace
Please make sure your IDE is adjusted to "trim right" on modified lines at least.

src/Model/Behavior/SearchBehavior.php
JayWalker512: Ah you are right, that way is better. Forgive my thick-headedness :)

src/Model/Behavior/SearchBehavior.php
```diff
@@ -159,22 +159,29 @@ protected function _extractParams($params, $filters)
...
-            if (is_array($value)) {
-                foreach ($value as $childKey => $childValue) {
-                    if (!is_numeric($childKey)) {
-                        $flattened[$key . '.' . $childKey] = $childValue;
-                    } else {
-                        $flattened[$key][$childKey] = $childValue;
-                    }
-                }
-            } else {
+            if (is_array($value) && !empty($filters[$key]) && $filters[$key]->getConfig()['flatten'] == false) {
```
ADmad: Use `===`.

dereuromark:
This looks good but is probably missing a small test case to prevent regressions here in the future.

JayWalker512:
> This looks good but is probably missing a small test case to prevent regressions here in the future.
Yes, what is the best way to go about that? I was having trouble getting PHPUnit to cooperate with the tests locally, would it be better for me to just write a test and let Travis CI deal with it after commit?

ADmad:
@JayWalker512 Given that we can only review the test results on travis and not your PC sure :smile:

JayWalker512:
@ADmad Of course, I just meant for my own purposes of ensuring the test is sensible & works, rather than piling commits here and hoping for the best. I'll take a look at the other tests and see what happens.

ADmad:
@JayWalker512  I see phpunit hasn't been added as a dev dependency currently. So  just run `composer install --dev phpunit/phpunit:"<6.0"` and then you should be able to run `vendor/bin/phpunit` without issues.

Also later on you can squash the commits into a single commit.

JayWalker512:
How are these?

josegonzalez:
@ADmad *bump*

src/Model/Behavior/SearchBehavior.php
```diff
@@ -159,22 +159,29 @@ protected function _extractParams($params, $filters)
...
-            if (is_array($value)) {
-                foreach ($value as $childKey => $childValue) {
-                    if (!is_numeric($childKey)) {
-                        $flattened[$key . '.' . $childKey] = $childValue;
-                    } else {
-                        $flattened[$key][$childKey] = $childValue;
-                    }
-                }
-            } else {
+            if (is_array($value) && !empty($filters[$key]) && $filters[$key]->getConfig()['flatten'] === false) {
```
ADmad: Please change `->getConfig()` to `config()`, this plugin supports cake 3.0+.

src/Model/Behavior/SearchBehavior.php
```diff
@@ -159,22 +159,29 @@ protected function _extractParams($params, $filters)
...
+                $flattened[$key] = $value;
```
ADmad: Hmm.. Codecov says these 2 lines don't have test coverage.

tests/TestCase/Model/Behavior/SearchBehaviorTest.php
```diff
@@ -68,23 +68,28 @@ public function testProcessFilters()
         $behavior = $this
             ->getMockBuilder('Search\Model\Behavior\SearchBehavior')
             ->setConstructorArgs([$this->Comments])
-            ->setMethods(['_getAllFilters'])
+            ->setMethods(['_getAllFilters', '_flattenParams'])
```
ADmad: LOL, you mocked the method you made changes too allowing you to assert any input and output you want. Lines 170, 171 which are the main change you made are never really executed by the test suite.

ADmad:
I took care of changes required. Thanks for the PR.

PR #167, opened 2017-04-29
Remove deprecations for new major.

dereuromark:
To be merged after the other 3 PRs - and before tagging 3.0.1.

PR #175, opened 2017-08-04
Next

ADmad:
This is for next major release. Notable changes:

- Core dep. bumped to 3.4.
- All deprecated code removed.
- Ability to have `Table::searchConfiguration()` method removed. I have documented how to achieve the same separation of code by adding `Table::searchManager()` instead.
- Removed `queryStringToData` option of `PrgComponent`. Setting `$this->request->data` was pretty hackish. I have documented how the search inputs can still be easily populated from query string.


README.md
```diff
@@ -95,26 +87,26 @@ You can use `SearchManager::add()` method to add filter or use specific methods
...
 
-    public function searchConfiguration()
+        // Add the behaviour to your table
+        $this->addBehavior('Search.Search');
+    }
+
+    public function searchManager()
     {
-        $search = new Manager($this);
+        $searchManager = $this->behaviors()->Search->searchManager();
```
dereuromark: I believe you can also directly do 

    $this->searchManager()

This is typehinted in the IDE due to the IdeHelper and the loaded behavior :)

dereuromark:
Looks good to be on a first glance.

dereuromark:
Will you include https://github.com/FriendsOfCake/search/pull/167  for the release?

README.md
ADmad: Calling $this->searchManager() inside Table::searchManager() will cause an infinite loop :)

dereuromark: Yeah this is a neat way of overwriting the parent "call" it seems.

dereuromark: Small downside only is that we lose some IDE clickability here probably, compared to the previous way.

ADmad:
@burzum You too use this plugin right? Let me know if you have any feedback on this.

burzum:
@ADmad seems to be OK for me! 👍

PR #179, opened 2017-08-19
Call searchManager() through table instance.

### LINKED ISSUE #176: Documented searchManager config does not work
Does the "overwriting of searchManager" work for you?
This didnt for me:
```php
	/**
	 * @return \Search\Manager
	 */
	public function searchManager() {
		/** @var \Search\Manager $searchManager */
		$searchManager = $this->behaviors()->Search->searchManager();

		$searchManager
			->value('type');

		return $searchManager;
```
Besides the issue that it needs the typehint, this is not even invoked for me.

I had to fix it using my own method:
```php
	/**
	 * @param array $config The configuration for the Table.
	 * @return void
	 */
	public function initialize(array $config) {
		parent::initialize($config);

		$this->addBehavior('Search.Search');
		$this->searchConfiguration($this->searchManager());
	}

	/**
	 * @param \Search\Manager
	 *
	 * @return void
	 */
	public function searchConfiguration(Manager $searchManager) {
		$searchManager
			->value('type');
	}
```

If that is to be confirmed we should fix the readme docs.

I would vote otherwise to have the searchConfiguration callback back, as it simplified this one line more.
And one major BC break less to fix up (if you have a lot of searched models this is quite painful).

ADmad:
Closes #176

### DIFF
--- a/src/Model/Behavior/SearchBehavior.php
+++ b/src/Model/Behavior/SearchBehavior.php
@@ -174,7 +174,7 @@ class SearchBehavior extends Behavior
      */
     protected function _getFilters($collection = 'default')
     {
-        return $this->searchManager()->getFilters($collection);
+        return $this->_table->searchManager()->getFilters($collection);
     }
 
     /**


ADmad:
@dereuromark Can you please test and confirm this patch fixes the issue?

dereuromark:
👍 Works

PR #183, opened 2017-09-26
Because magic is evil

davidyell:
Replace the use of an inferred magic property for a concrete getter

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ You can use `SearchManager::add()` method to add filter or use specific methods
 like `value()`, `like()` etc. for in built filters.
 
 If you do not want to clutter your `initialize()` method with search config you
-can instead add a `searchManager()` method to the table class and reutrn a search
+can instead add a `searchManager()` method to the table class and return a search
 manager instance.
 
 ```php
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ class ExampleTable extends Table
     public function searchManager()
     {
         /** @var \Search\Manager $searchManager */
-        $searchManager = $this->behaviors()->Search->searchManager();
+        $searchManager = $this->behaviors()->get('Search')->searchManager();
         $searchManager
             ->like('title')
             ->value('status');


README.md
```diff
@@ -118,7 +118,7 @@ class ExampleTable extends Table
     public function searchManager()
     {
         /** @var \Search\Manager $searchManager */
-        $searchManager = $this->behaviors()->Search->searchManager();
+        $searchManager = $this->behaviors()->get('Search')->searchManager();
```
dereuromark: Do we have typehinting or autocomplete on the getter here? Otherwise this is no improvement at all. Magic strings are usually worse then the attributes.
I would be fine with using get(magicstring) + IdeHelper generate of Behaviors::get() etc.

davidyell: Well my motivation to change it was because PHP Storm underlined it as a warning for a magic property.

dereuromark: We need to annotate/typehint on the behaviors() collection then, using the meta file etc.
That should fix it for IDEs?

dereuromark: Hm, I just checked.. The only way to get autocomplete working for
```php
$searchManager = $this->behaviors()->get('Search')->searchManager();
```

Is to use the following meta map on the ObjectRegistry (unfortunately, the BehaviorRegistry doesnt work):
```php
	override(
		\Cake\Core\ObjectRegistry::get(0),
		map([
			'Search' => \Search\Model\Behavior\SearchBehavior::class,
		])
	);
```

inoas:
```diff
-        $searchManager = $this->behaviors()->Search->searchManager();
+        $searchManager = $this->behaviors()->get('Search')->searchManager();
```

Note that this is probably also slower.

dereuromark: Speed is not relevant here imo.
the important thing missing is ide comp.

inoas: That's the attitude that I am concerned about (not this special case anyway, not a problem).

dereuromark: :) Lets talk on slack if you dont understand the difference on this suspect. Please dont shift focus away from what actually matters here.
Your "concerned about" remark was also not neccessary :/

dereuromark: @davidyell Back to topic

I got actually the property to work, the magic string getter not..
How about we start using
```php
$searchManager = $this->behaviors()->search()->searchManager();
```
etc?
That would also work to be autocompleted/discovered.

dereuromark:
In light of https://github.com/dereuromark/cakephp-ide-helper/pull/56 I would rather see this reverted.
Because the actual evil here is magic string usage.

Issue #191, opened 2017-10-17
This will remove boolean values


dilab:
https://github.com/FriendsOfCake/search/blob/a836be3000be12aadcb75c6305b85ac278e89a78/src/Model/Behavior/SearchBehavior.php#L104

The problematic piece is `Hash::filter($params)`, it will remove pair with **false** as its value. Which is not desired, because $params['key'] = false is a valid filter value

ADmad:
Normally request's query string is passed as search params and hence are always going to be string values.

inoas:
So using search for, say, a JSON API is out of question?

ADmad:
A patch to allow valid non empty, non string values is welcome.

dilab:
@inoas same here. I am using search in an API end point.

dilab:
@ADmad PR done at https://github.com/FriendsOfCake/search/pull/193

inoas:
This should get you going if you want to start a PR just allowing `false`:

```php
$data = ['nested' =>
    false,
    null,
    '0',
    '0.0',
    0,
    0.0,
    true,
    ' ',
]];

$filterFn = function($var) {
    return $var === 0 || $var === 0.0 || $var === '0' || $var === false || $var === null || !empty($var);
};

$res = \Cake\Utility\Hash::filter($data, $filterFn);
var_dump($res);
```

However I assume then filter doesn't make any sense if false AND null are allowed.

@ADmad would it be okay to make searchManager not filter at all? `array_intersect_key()` should be good enough/the relevant thing.

If not, would it be a good option (then disabled by default)?

https://github.com/FriendsOfCake/search/blob/a836be3000be12aadcb75c6305b85ac278e89a78/src/Model/Behavior/SearchBehavior.php#L64

Would become an option (array_intersect would happen anyway).

dereuromark:
@inoas You should probably comment on the existing PR: https://github.com/FriendsOfCake/search/pull/193

PR #193, opened 2017-10-17
allow non empty string values

dilab:
Fix for https://github.com/FriendsOfCake/search/issues/191

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,16 @@ class ExampleTable extends Table
 }
 ```
 
+#### Empty Values
+By default, `['', false, null]` are treated as empty values and will be filtered out. If you wish to
+alter this behavior, you can overwrite the values using `emptyValues` key.
+
+```php
+$this->addBehavior('Search.Search', [
+    'emptyValues' => ['']
+]);
+```
+
 ### Controller class
 In order for the Search plugin to work it will need to process the query params
 which are passed in your URL. So you will need to edit your `index` method to
--- a/src/Model/Behavior/SearchBehavior.php
+++ b/src/Model/Behavior/SearchBehavior.php
@@ -121,6 +121,7 @@ class SearchBehavior extends Behavior
     {
         return array_intersect_key(Hash::filter($params, function ($val) {
             $emptyValues = (array)$this->getConfig('emptyValues');
+
             return !in_array($val, $emptyValues, true);
         }), $filters);
     }
--- a/src/Model/Behavior/SearchBehavior.php
+++ b/src/Model/Behavior/SearchBehavior.php
@@ -31,6 +31,7 @@ class SearchBehavior extends Behavior
             'searchManager' => 'searchManager',
             'isSearch' => 'isSearch'
         ],
+        'emptyValues' => ['']
     ];
 
     /**
@@ -103,7 +104,12 @@ class SearchBehavior extends Behavior
     protected function _extractParams($params, $filters)
     {
         return array_intersect_key(Hash::filter($params, function ($val) {
-            return $val !== '' && $val !== null;
+            foreach ($this->getConfig('emptyValues') as $emptyValue) {
+                if ($val === $emptyValue) {
+                    return false;
+                }
+            }
+            return true;
         }), $filters);
     }
 
--- a/src/Model/Behavior/SearchBehavior.php
+++ b/src/Model/Behavior/SearchBehavior.php
@@ -53,7 +53,7 @@ class SearchBehavior extends Behavior
         parent::initialize($config);
 
         if (isset($config['emptyValues'])) {
-            $this->_defaultConfig['emptyValues'] = $config['emptyValues'];
+            $this->setConfig('emptyValues', $config['emptyValues'], false);
         }
     }
 
@@ -120,7 +120,7 @@ class SearchBehavior extends Behavior
     protected function _extractParams($params, $filters)
     {
         return array_intersect_key(Hash::filter($params, function ($val) {
-            return !in_array($val, $this->_defaultConfig['emptyValues'], true);
+            return !in_array($val, $this->getConfig('emptyValues'), true);
         }), $filters);
     }
 
--- a/src/Model/Behavior/SearchBehavior.php
+++ b/src/Model/Behavior/SearchBehavior.php
@@ -45,7 +45,8 @@ class SearchBehavior extends Behavior
     /**
      * Overwrite emptyValues config value
      *
-     * @param array $config
+     * @param array $config Config
+     * @return void
      */
     public function initialize(array $config)
     {
--- a/src/Model/Behavior/SearchBehavior.php
+++ b/src/Model/Behavior/SearchBehavior.php
@@ -22,7 +22,11 @@ class SearchBehavior extends Behavior
     /**
      * Default config for the behavior.
      *
+     * You can overwrite default empty values using emptyValues key
+     * when initialise the behavior
+     *
      * @var array
+     *
      */
     protected $_defaultConfig = [
         'implementedFinders' => [
--- a/src/Model/Behavior/SearchBehavior.php
+++ b/src/Model/Behavior/SearchBehavior.php
@@ -123,9 +123,9 @@ class SearchBehavior extends Behavior
      */
     protected function _extractParams($params, $filters)
     {
-        return array_intersect_key(Hash::filter($params, function ($val) {
-            $emptyValues = (array)$this->getConfig('emptyValues');
+        $emptyValues = (array)$this->getConfig('emptyValues');
 
+        return array_intersect_key(Hash::filter($params, function ($val) use ($emptyValues) {
             return !in_array($val, $emptyValues, true);
         }), $filters);
     }
--- a/src/Model/Behavior/SearchBehavior.php
+++ b/src/Model/Behavior/SearchBehavior.php
@@ -120,7 +120,8 @@ class SearchBehavior extends Behavior
     protected function _extractParams($params, $filters)
     {
         return array_intersect_key(Hash::filter($params, function ($val) {
-            return !in_array($val, $this->getConfig('emptyValues'), true);
+            $emptyValues = (array)$this->getConfig('emptyValues');
+            return !in_array($val, $emptyValues, true);
         }), $filters);
     }
 
--- a/src/Model/Behavior/SearchBehavior.php
+++ b/src/Model/Behavior/SearchBehavior.php
@@ -42,6 +42,15 @@ class SearchBehavior extends Behavior
      */
     protected $_isSearch = false;
 
+    public function initialize(array $config)
+    {
+        parent::initialize($config);
+
+        if (isset($config['emptyValues'])) {
+            $this->_defaultConfig['emptyValues'] = $config['emptyValues'];
+        }
+    }
+
     /**
      * Callback fired from the controller.
      *
@@ -105,7 +114,7 @@ class SearchBehavior extends Behavior
     protected function _extractParams($params, $filters)
     {
         return array_intersect_key(Hash::filter($params, function ($val) {
-            return !in_array($val, $this->getConfig('emptyValues'), true);
+            return !in_array($val, $this->_defaultConfig['emptyValues'], true);
         }), $filters);
     }
 
--- a/src/Model/Behavior/SearchBehavior.php
+++ b/src/Model/Behavior/SearchBehavior.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace Search\Model\Behavior;
 
 use Cake\ORM\Behavior;
@@ -92,7 +93,7 @@ class SearchBehavior extends Behavior
     }
 
     /**
-     * Extracts all parameters for wich a filter with a matching field
+     * Extracts all parameters for which a filter with a matching field
      * name exists.
      *
      * @param array $params The parameters array to extract from.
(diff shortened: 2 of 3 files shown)


dereuromark:
Looks good to me.

inoas:
For us `null` is a valid JSON value. Even searching for `''` can make a lot of sense.

What about making the filter function optional instead of adding yet more and more values that could, could not be filterable, depending on app use case.

https://github.com/FriendsOfCake/search/blob/a836be3000be12aadcb75c6305b85ac278e89a78/src/Model/Behavior/SearchBehavior.php#L64

Would become an option (array_intersect would happen anyway).

dereuromark:
@inoas Isnt this patch already better than the current behavior which filters out all falsey values?
I think the proposed change here should be the "default", but I do agree that we could make it configurable as array map of characters that would be valid and should not be removed.

inoas:
The current behavior just filters out `false` not `'false'` afaik. Anyway, yes this change is better 👍 so even if no/small say: LGTM.

dereuromark:
Well, 'false' never was a valid value to begin with :) You would need custom logic on project level for that either way :)

inoas:
Sure but `false` would be filtered and now you can search data you could not before.

ADmad:
While this works I think it would be better to just make the list of filtered values to be configurable, so that we don't have to change code again in future where someone wants to use empty string or null too as valid values.

ADmad:
The PrgComponent has an `emptyValues` options. Same could be add to behavior to with default mimicking existing behavior.

dilab:
Just want to check if we are merging this PR or not?

dereuromark:
@dilab Yes, but it needs the adjustment mentioned to have the list configurable.

dilab:
@dereuromark @AD7six @inoas How do you guys want to pass in the config option.
Will something like below work for you?

```
$this->Articles->find('search', ['search' => $queryString, 'validValues'=>[null,false]);
```

ADmad:
The option would be used during behaviour config, not passed in find call.

P.S. Pay more attention to the user names. It's ADmad not AD7six ;)

dilab:
@ADmad  
Ooops. sorry about that.

Will this do?

```
$this->addBehavior('Search.Search','validValues'=>[null,false]);
```

ADmad:
Instead of `validValues` use key `emptyValues` to indicate which values should be treated as empty and filtered out. It also maintains consistency with the PrgComponent's option.

dilab:
@ADmad  added **emptyValues** config

tests/TestCase/Model/Behavior/SearchBehaviorTest.php
```diff
@@ -186,6 +187,32 @@ public function testFinder()
...
+            'emptyValues' => ['a']
+        ]);
+        $query = $this->Articles->find('search', ['search' => $queryString]);
+        $this->assertEquals(2, $query->clause('where')->count());
+
+        $this->Articles->behaviors()->get('Search')->config([
+            'emptyValues' => ['a', false]
+        ]);
+        $query = $this->Articles->find('search', ['search' => $queryString]);
+        $this->assertEquals(1, $query->clause('where')->count());
```
dereuromark: assertSame() please to avoid type issues.

src/Model/Behavior/SearchBehavior.php
```diff
@@ -101,7 +103,14 @@ public function searchManager()
      */
     protected function _extractParams($params, $filters)
     {
-        return array_intersect_key(Hash::filter($params), $filters);
+        return array_intersect_key(Hash::filter($params, function ($val) {
+            foreach ($this->getConfig('emptyValues') as $emptyValue) {
+                if ($val === $emptyValue) {
```
dereuromark: What about null? Was that filtered out so far, then emptyValues default above needs to also contain that one for BC.

src/Model/Behavior/SearchBehavior.php
```diff
@@ -101,7 +103,14 @@ public function searchManager()
...
+                    return false;
+                }
+            }
+            return true;
```
inoas: New line is missing between same-scope-block and return.

src/Model/Behavior/SearchBehavior.php
dilab: For BC, I can add following to the default config:
```
'emptyValues' => ['',null]
```

What else should I add?

tests/TestCase/Model/Behavior/SearchBehaviorTest.php
dilab: hmm, I am keeping it consistent with other tests. 
they are all using **assertEquals**

src/Model/Behavior/SearchBehavior.php
inoas: Before: https://api.cakephp.org/3.0/source-class-Cake.Utility.Hash.html#618 (3.0)
https://api.cakephp.org/3.5/source-class-Cake.Utility.Hash.html#643-652 (3.5)

Aka: if `($var === 0 || $var === '0' || !empty($var))` (3.0) and `0.0` (3.5) https://api.cakephp.org/3.5/source-class-Cake.Utility.Hash.html#643-652

For empty() see: https://3v4l.org/sj8bM

So all these "generally" need to be filtered by default:

```php
    '',
    null,
    false,
    0,
    0.0,
    '0',
```

Minus the exception of `Hash::filter()`'s defaults: `$var === 0 || $var === '0'`
And since some time `0.0` https://api.cakephp.org/3.5/source-class-Cake.Utility.Hash.html#643-652

Leaves these as a requirement to be filtered out by `'emptyValues'` by convention/default:

```php
    '',
    null,
    false,
```

p.s.: We should also have an option to ltrim/rtrim inputs if it is not there already.

tests/TestCase/Model/Behavior/SearchBehaviorTest.php
dereuromark: No excuse to do it right IMO :)

src/Model/Behavior/SearchBehavior.php
dereuromark: Just make sure it behaves the same by default as the current Hash::filter() to avoid BC issues.

src/Model/Behavior/SearchBehavior.php
dereuromark: newline missing above to please CS

src/Model/Behavior/SearchBehavior.php
```diff
@@ -30,6 +31,7 @@ class SearchBehavior extends Behavior
             'searchManager' => 'searchManager',
             'isSearch' => 'isSearch'
         ],
+        'emptyValues' => ['']
```
inoas: This could have one of these key names instead:

- `'filterValues'`
- `'discardValues'`
- `'ignoreValues'`

Then it could make sense to supply a callable from your app and filter arbitrary values.

src/Model/Behavior/SearchBehavior.php
inoas: Not sure why my comment appears twice.

src/Model/Behavior/SearchBehavior.php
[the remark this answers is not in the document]
ADmad: Also you will have to ensure the array gets overwritten and not merged (which is default behavior) when custom value is set otherwise one would never be able to remove the default empty values.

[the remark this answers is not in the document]
inoas: If we support CakePHP 3.3 we could add '0.0'.
- `Hash::filter()`: https://api.cakephp.org/3.3/source-class-Cake.Utility.Hash.html#645
- `empty()`: https://3v4l.org/sj8bM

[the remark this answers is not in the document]
dilab: > Also you will have to ensure the array gets overwritten and not merged (which is default behavior) 
@ADmad actually how can I do that? I am not sure this part.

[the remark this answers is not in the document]
dilab: @ADmad I realise that users can use configShallow to overwrite the config values

tests/TestCase/Model/Behavior/SearchBehaviorTest.php
```diff
@@ -186,6 +187,35 @@ public function testFinder()
...
+        $this->assertSame(2, $query->clause('where')->count());
+
+        $this->Articles->behaviors()->get('Search')->configShallow([
+            'emptyValues' => ['a']
+        ]);
+        $query = $this->Articles->find('search', ['search' => $queryString]);
+        $this->assertSame(2, $query->clause('where')->count());
+
+        $this->Articles->behaviors()->get('Search')->configShallow([
+            'emptyValues' => ['a', false]
```
ADmad: This test is pointless. The `emptyValues` should be overwritten even when initially setting up the behavior using `addBehavior()` which doesn't happen currenlty. For that you will have to add `SearchBehavior::initialize()` and overwrite `emptyValues` if present in passed `$config` array.

dilab:
@ADmad change is made as requested.

src/Model/Behavior/SearchBehavior.php
```diff
@@ -42,6 +42,15 @@ class SearchBehavior extends Behavior
      */
     protected $_isSearch = false;
 
+    public function initialize(array $config)
+    {
+        parent::initialize($config);
+
+        if (isset($config['emptyValues'])) {
+            $this->_defaultConfig['emptyValues'] = $config['emptyValues'];
```
ADmad: You  need to set the key in `_config` (through `config()` method with `merge` false) not in `_defaultConfig`.

dilab:
@ADmad changes made

src/Model/Behavior/SearchBehavior.php
```diff
@@ -120,7 +120,7 @@ public function searchManager()
     protected function _extractParams($params, $filters)
     {
         return array_intersect_key(Hash::filter($params, function ($val) {
-            return !in_array($val, $this->_defaultConfig['emptyValues'], true);
+            return !in_array($val, $this->getConfig('emptyValues'), true);
```
dereuromark: Can we trust this to be always an array?
Otherwise I would store in a local `$emptyValues = (array)...` with a cast first.

dilab: Good point

inoas:
> Instead of validValues use key emptyValues to indicate which values should be treated as empty and filtered out. It also maintains consistency with the PrgComponent's option.

So any comments on using yet another term: `filterValues` or `filteredValues`, cause that is really what is happening?

dilab:
@inoas I think we have already settled on **emptyValues** ->https://github.com/FriendsOfCake/search/pull/193#issuecomment-337802929

inoas:
Well that's another thing validValues was opt-in, filtered/emptyValues is opt-out.
`emptyValues` could very well contain more than null-ish values in apps, so IMHO the variable name is kind of lying compared to what it is/can be used for. What does Admad say?

dilab:
@ADmad @dereuromark changes made

src/Model/Behavior/SearchBehavior.php
```diff
@@ -101,7 +119,11 @@ public function searchManager()
      */
     protected function _extractParams($params, $filters)
     {
-        return array_intersect_key(Hash::filter($params), $filters);
+        return array_intersect_key(Hash::filter($params, function ($val) {
+            $emptyValues = (array)$this->getConfig('emptyValues');
```
ADmad: Move this statement out of the closure and pass `emptyValues` as  `function ($val) use ($emptyValues) {` to avoid making the same method call on each iteration.

src/Model/Behavior/SearchBehavior.php
```diff
@@ -30,6 +32,7 @@ class SearchBehavior extends Behavior
             'searchManager' => 'searchManager',
             'isSearch' => 'isSearch'
         ],
+        'emptyValues' => ['', false, null]
```
ADmad: This new option needs to be documented in docblock above and also in readme.

dilab:
@ADmad changes made

README.md
```diff
@@ -127,6 +127,16 @@ class ExampleTable extends Table
 }
 ```
 
+#### Empty Values
+By default, `['', false, null]` are treated as empty values and will be filtered out. If you wish to
+alter this behavior, you can overwrite the values using `emptyValues` key.
+
+```php
+$this->addBehavior('Search.Search', [
+    'emptyValues' => ['']
```
dereuromark: Does this actually work? Or do we need explicit shallow merging here to overwrite because CakePHP uses the wrong merge operation by default?

src/Model/Behavior/SearchBehavior.php
```diff
@@ -22,7 +22,11 @@ class SearchBehavior extends Behavior
     /**
      * Default config for the behavior.
      *
+     * You can overwrite default empty values using emptyValues key
+     * when initialise the behavior
```
dereuromark: initializating

inoas: Proper English is very posh 💛 :trollface:

dereuromark: Sry :) Typo fixed.

dilab: lol. done

README.md
dilab: This will work. When emptyValues is supplied when initializing, the shallow merge is used to overwrite emptyValues

dilab: Reference: https://github.com/dilab/search/blob/master/tests/TestCase/Model/Behavior/SearchBehaviorTest.php#L206

ADmad: @dereuromark Check the behavior's `initialize()` method. I ensured he made required change so that `emptyValues` key is overwritten not merged :slightly_smiling_face:

dilab:
@ADmad do you plan to merge my changes anytime soon?

src/Model/Behavior/SearchBehavior.php
dilab: Already documented

PR #198, opened 2017-11-17
Allow _isSearch auto populate

dereuromark:
So far you always had to manually add this in every single controller:
```php
    $this->set('_isSearch', $this->Capabilities->isSearch());
```
To allow for 
```php
    if (!empty($_isSearch)) {
        echo $this->Html->link('Reset', ['action' => 'index'], ['class' => 'button']);
    }
```

With the component easily being able to do that by default for the conventional way, this will simplify quite a bit the code here.
You can also configure it, or even disable it completely by passing e.g. false.

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,8 +214,9 @@ The array passed to `FormHelper::create()` will cause the helper to create an
 query params.
 
 #### Adding a reset button dynamically
-Since the Prg component will pass down the information on whether the query was modified by your
-search query string, you can also include a reset button only if necessary:
+The Prg component will pass down the information on whether the query was modified by your
+search query string by setting `$_isSearch` view variable to true here in this case.
+This way you can include a reset button only if necessary:
 ```php
 // in your form
 if (!empty($_isSearch)) {
--- a/README.md
+++ b/README.md
@@ -214,8 +214,8 @@ The array passed to `FormHelper::create()` will cause the helper to create an
 query params.
 
 #### Adding a reset button dynamically
-Since the Prg component can pass down the information on whether the query was modified by your
-search query strings, you can also include a reset button only if necessary:
+Since the Prg component will pass down the information on whether the query was modified by your
+search query string, you can also include a reset button only if necessary:
 ```php
 // in your form
 if (!empty($_isSearch)) {
--- a/README.md
+++ b/README.md
@@ -214,18 +214,11 @@ The array passed to `FormHelper::create()` will cause the helper to create an
 query params.
 
 #### Adding a reset button dynamically
-If you pass down the information on whether the query was modified by your
+Since the Prg component can pass down the information on whether the query was modified by your
 search query strings, you can also include a reset button only if necessary:
-
-```php
-// in your controller action
-$this->set('isSearch', $this->Articles->isSearch());
-```
-
-Then one can switch based on that in the template:
 ```php
 // in your form
-if ($isSearch) {
+if (!empty($_isSearch)) {
     echo $this->Html->link('Reset', ['action' => 'index']);
 }
 ```
--- a/src/Controller/Component/PrgComponent.php
+++ b/src/Controller/Component/PrgComponent.php
@@ -20,6 +20,7 @@ class PrgComponent extends Component
      *   (will not be passed along in the URL).
      * - `modelClass` : Configure the controller's modelClass to be used for the query, used to
      *   populate the _isSearch view variable to allow for a reset button, for example.
+     *   Set to false to disable the auto-setting of the view variable.
      *
      * @var array
      */
--- a/src/Controller/Component/PrgComponent.php
+++ b/src/Controller/Component/PrgComponent.php
@@ -63,18 +63,23 @@ class PrgComponent extends Component
      */
     public function beforeRender()
     {
+        if (!$this->_actionCheck()) {
+            return null;
+        }
+
         $controller = $this->_registry->getController();
         $modelClass = $this->getConfig('modelClass', $controller->modelClass);
         if (!$modelClass) {
             return null;
         }
 
-        if (!isset($controller->{$modelClass})) {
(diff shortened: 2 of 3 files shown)


src/Controller/Component/PrgComponent.php
```diff
@@ -53,6 +53,35 @@ public function startup()
...
+     */
+    public function beforeRender()
+    {
+        $controller = $this->_registry->getController();
+        $modelClass = $this->getConfig('modelClass', $controller->modelClass);
+        if (!$modelClass) {
+            return null;
+        }
+
+        if (!isset($controller->{$modelClass})) {
```
ADmad: For plugins value of `$controller->modelClass` will be of form `Foo.Bar` so this check will fail. You'll need to use `pluginSplit()` first.

src/Controller/Component/PrgComponent.php
```diff
@@ -53,6 +53,35 @@ public function startup()
...
+     * Populates the $_isSearch view variable based on the current request.
+     *
+     * You need to configure the modelClass config if you are not using the controller's
+     * default modelClass property.
+     *
+     * @return \Cake\Http\Response|null
+     */
+    public function beforeRender()
+    {
+        $controller = $this->_registry->getController();
```
ADmad: You should first check if current action is in list on configured `actions` and bail out early. Current this method will unnecessarily run for all actions.

src/Controller/Component/PrgComponent.php
dereuromark: Ah right, thanks.

ADmad: Actually an even better way to get model instance would be using:
```php
$modelClass = $this->getConfig('model');
$model = $controller->loadModel($modelClass);
```
If `$modelClass` is `null` the controller's default model will be automatically returned.

dereuromark: If the model at this point is not loaded yet, I assume it should not be loaded either (as the value returned must be logically false anyway - as no search query has been done at all).

dereuromark:
I also added 2 tests.

src/Controller/Component/PrgComponent.php
ADmad: If `$modelClass` is same as `Controller::$modelClass`, which is going to be the case most of the times, `$controller->{$modelClass}` is always gonna trigger model loading. But okay we can keep current code.

README.md
```diff
@@ -214,18 +214,11 @@ The array passed to `FormHelper::create()` will cause the helper to create an
...
-If you pass down the information on whether the query was modified by your
-search query strings, you can also include a reset button only if necessary:
-
-```php
-// in your controller action
-$this->set('isSearch', $this->Articles->isSearch());
-```
-
-Then one can switch based on that in the template:
+Since the Prg component will pass down the information on whether the query was modified by your
```
ADmad: It would be better if this statement was reworded to explicitly mention that `$_isSearch` view var is auto set if a search has been performed. This would make is more clear for newbies.

src/Controller/Component/PrgComponent.php
```diff
@@ -21,6 +18,8 @@ class PrgComponent extends Component
      * - `queryStringBlacklist` : An array of form fields that should not end up in the query.
      * - `emptyValues` : A map of fields and their values to be considered empty
      *   (will not be passed along in the URL).
+     * - `modelClass` : Configure the controller's modelClass to be used for the query, used to
+     *   populate the _isSearch view variable to allow for a reset button, for example.
```
ADmad: Please add that setting this config to `false` will disable the auto setting of view var.

dereuromark: Done.

src/Controller/Component/PrgComponent.php
```diff
@@ -53,6 +54,40 @@ public function startup()
...
+     * Populates the $_isSearch view variable based on the current request.
+     *
+     * You need to configure the modelClass config if you are not using the controller's
+     * default modelClass property.
+     *
+     * @return \Cake\Http\Response|null
+     */
+    public function beforeRender()
+    {
+        if (!$this->_actionCheck()) {
```
ADmad: I realized we should also bail out if response type is not `text/html`, else setting the view var could inadvertently modify API response.

dereuromark: Fair point.

dereuromark: Actually, thinking about it some more: Wouldn't this modify the response in a good way, just adding the additional info here?
I mean, if one involves the Prg component, then probably not for a stateless, but somewhat mixed or stateful API, and here one would probably also want to act on the info on whether the request contains a search filter or not.
What's the harm here in letting it add this on top by default?

ADmad: The same controller could be used for html as well as json responses. Harmful or not unexpected/unwanted changes in output is a bad thing.

ADmad: The same controller could be used for html as well as json responses. Harmful or not unexpected/unwanted changes in output is a bad thing.

dereuromark:
I just tried it with .json and a default baked view as

        ...
        $this->set(compact('capabilities'));
        $this->set('_serialize', ['capabilities']);

And it is not added by default due to the explicit setting here (as recommended anyway).
Thus there is no issue for serializing in json or xml etc, and can however easily be added on top if desired.

If you don't mind, I would like to keep the code this way, as it is easier to opt-in that part if needed, rather then trying to disable the disabling here.

ADmad:
Okay, I guess we don't have to check for non HTML responses in that case. 

But instead of setting `model` option to `false` to disable this feature I would like the option of disabling the triggering of beforeRender handler itself. I can implement that if you agree.

dereuromark:
Can we merge to master and have a 2nd pr here? We just dont tag until then.

PR #203, opened 2017-12-30
Add support for searching other datasources

Marlinc:
This change adds a trait which can be used in custom repository
classes and adds support for custom repositories to the value and
boolean filters.

Signed-off-by: Marlin Cremers <m.cremers@cvo-technologies.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,36 @@ public function initialize()
 The `Search.Prg` component will allow your filtering forms to be populated using
 the data in the query params. It uses the [Post, redirect, get pattern](https://en.wikipedia.org/wiki/Post/Redirect/Get).
 
+### Custom repository
+
+It is also possible to use the search plugin on custom repositories which
+implement `Cake\Datasource\RepositoryInterface` like endpoint classes used
+in the Webservice plugin.
+
+```php
+<?php
+
+namespace App\Model\Endpoint;
+
+use Muffin\Webservice\Model\Endpoint;
+use Search\Model\SearchTrait;
+
+class ProductsEndpoint extends Endpoint
+{
+    use SearchTrait;
+
+    public function initialize()
+    {
+        $this->searchManager()
+            ->value('category_id');
+    }
+}
+
+```
+
+After including the trait you can use the searchManager by calling the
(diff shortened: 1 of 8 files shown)


ADmad:
Failing test seems related to this change.

ADmad:
@Marlinc ping

Marlinc:
Working now @ADmad

src/Model/SearchTrait.php
```diff
@@ -0,0 +1,215 @@
...
+use Exception;
+use Search\Manager;
+
+trait SearchTrait
+{
+
+    /**
+     * Search Manager instance.
+     *
+     * @var \Search\Manager
```
dereuromark: Does this need |null ?

Marlinc: Add it

src/Model/Filter/Value.php
```diff
@@ -27,6 +28,16 @@ public function process()
             return false;
         }
 
+        if (!$this->manager()->getRepository() instanceof Table) {
```
ADmad: Shouldn't this block go after the multi values check below?

ADmad:
Would be nice if some info was added to the readme about these changes.

Marlinc:
I've added a small text to the README, I hope it will be clear to users

src/Model/Filter/Value.php
Marlinc: Let me check

Marlinc: Changed :smile:

README.md
```diff
@@ -182,6 +182,35 @@ public function initialize()
 The `Search.Prg` component will allow your filtering forms to be populated using
 the data in the query params. It uses the [Post, redirect, get pattern](https://en.wikipedia.org/wiki/Post/Redirect/Get).
 
+### Custom repository
+
+It is also possible to use the search plugin on custom repositories like
```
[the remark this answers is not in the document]
Marlinc: Done

bravo-kernel:
Aha, finally understand what "other datasources" means, thanks !

> custom repositories which implement `Cake\Datasource\RepositoryInterface`

PR #204, opened 2018-01-19
Finder map

dereuromark:
Almost always custom finders require certain keys to be present. And you cannot change that, especially not for plugin finders.

Make custom finders get the right fields via map functionality.

Refs https://github.com/dereuromark/cakephp-tools/pull/200 which needs this functionality.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 vendor/
 phpunit.xml
 composer.lock
+.idea/
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,6 @@ before_script:
 
   - composer install
 
-  - if [[ $PHPCS = 1 ]]; then composer require cakephp/cakephp-codesniffer:"2.*"; fi
   - if [[ $PHPSTAN = 1 ]]; then composer require phpstan/phpstan; fi
 
   - if [[ $DB = 'mysql' ]]; then mysql -e 'CREATE DATABASE cakephp_test;'; fi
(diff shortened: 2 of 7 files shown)


ADmad:
Would be a good idea to have a boolean option (maybe `onlyMapped`) to pass only the mapped fields as finder options?

dereuromark:
I guess we dont need that since they all ignore overhead keys afaik. Didnt meet that use case that. :)

ADmad:
BTW your use case can also be achieved using callback filter:

```php
$searchManager
    ->finder('callback', function ($query, $args) {
        $query->find('bits', ['bits' => $args['status']]);
    });

**Comment by dereuromark** [comment]
Anything can be achieved using that one ;)
The idea was to use the existing explicit finder one here IMO - as this is the tailored one for finder calls that should be preferred if possible.

**Comment by dereuromark** [comment]
@ADmad Actually, I was thinking, maybe we should invert from=>to into to=>from?
as they keys must be unique, and it could be more likely that you reuse the same incoming value twice, as the finder needs two keys (but same value) for example.

finder(custom, [foo=>x, bar=>x]) would then work with map of
```
'foo' => 'x',
'bar' => 'x'
```
and the posted key `x`.

I can quickly refactor and adjust docs before we release.

ADmad:
Yup flipping the key value makes sense.

ADmad:
Rebasing the feature branch onto master might make stickler happy.

dereuromark:
Thx, that worked :)

PR #209, opened 2018-03-12
Closes #208

### LINKED ISSUE #208: CakePHP 3.6 deprecation warnings
Ist anyone working on Code Migration to avoid deprecation warnings popping Up when using this in 3.6.0-betaX? If appreciated I would try to fix These.

swiffer:
Resolves https://github.com/FriendsOfCake/search/issues/208

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,7 @@
+# Git will always convert line endings to LF on checkout.
+# You should use this for files that must keep LF endings, even on Windows.
+* text eol=lf
+
 # Remove files for archives generated using `git archive`
 .editorconfig export-ignore
 .gitattributes export-ignore
@@ -5,7 +9,3 @@
 phpunit.xml.dist export-ignore
 .travis.yml export-ignore
 tests export-ignore
-
-# Git will always convert line endings to LF on checkout.
-# You should use this for files that must keep LF endings, even on Windows.
-* text eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,3 +5,7 @@
 phpunit.xml.dist export-ignore
 .travis.yml export-ignore
 tests export-ignore
+
+# Git will always convert line endings to LF on checkout.
+# You should use this for files that must keep LF endings, even on Windows.
+* text eol=lf
--- a/composer.json
+++ b/composer.json
@@ -27,19 +27,19 @@
     },
     "require-dev": {
         "friendsofcake/cakephp-test-utilities": "dev-master",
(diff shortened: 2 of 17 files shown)


.gitattributes
```diff
@@ -5,3 +5,7 @@
 phpunit.xml.dist export-ignore
 .travis.yml export-ignore
 tests export-ignore
+
+# Git will always convert line endings to LF on checkout.
+# You should use this for files that must keep LF endings, even on Windows.
+* text eol=lf
```
dereuromark: :+1: 
We usually put that on the top though
before the export-ignore part.

dereuromark:
You need to also bump the constraint then:

    "cakephp/cakephp": "^3.4"

is not true anymore after this one.

.gitattributes
swiffer: done ;)

.gitattributes
```diff
@@ -1,3 +1,7 @@
...
```
dereuromark: what about the auto part? https://github.com/cakephp/cakephp/blob/master/.gitattributes#L3

swiffer:
i think it's still working on 3.4 at least in 3.5 code works without problems!

dereuromark:
To make sure we should add the same I added to my plugin test matrix: https://github.com/dereuromark/cakephp-geo/blob/master/.travis.yml#L27

dereuromark:
You need 5.7.14 there ar least for phpunit.

swiffer:
@dereuromark ok, I'm done ;)

dereuromark:
Nice work!

composer.json
```diff
@@ -17,17 +17,17 @@
...
-            "role":"Contributor",
-            "homepage":"https://github.com/admad"
+            "name": "ADmad",
+            "role": "Contributor",
+            "homepage": "https://github.com/admad"
         }
     ],
     "require": {
-        "cakephp/cakephp": "^3.4"
+        "cakephp/cakephp": "^3.5"
```
ADmad: The changes don't work with 3.4?

swiffer: Sadly getting Errors when running the Testsuite in 3.4.0

dereuromark:
This should be a minor release.
no bc breaks but min requirements raised.

ADmad:
Yup, it will be a minor release due to cake version bump. I'll merge when I am at a PC.

ADmad:
I have merged this in to a `cake-36` branch instead of `master` to prepare for 3.6. Will merge `cake-36` to master once CakePHP 3.6 is released.

PR #220, opened 2018-05-21
Update README.md

hassifmohd:
adding limit control inside the Creating your form (search filter)


ADmad:
Changing the per page limit is related to the paginator and has no direct relation to the search plugin.

PR #221, opened 2018-05-28
Adding a FilterInterface

burzum:
Adding an interface for filters.

### DIFF
--- a/src/Model/Filter/Base.php
+++ b/src/Model/Filter/Base.php
@@ -100,6 +100,7 @@ abstract class Base implements FilterInterface
     /**
      * Get the database field name.
      *
+     * @deprecated Use fields() instead.
      * @return string|array
      */
     public function field()
--- a/src/Model/Filter/Base.php
+++ b/src/Model/Filter/Base.php
@@ -9,7 +9,7 @@ use Search\Manager;
  * Base class for search type classes.
  *
  */
-abstract class Base
+abstract class Base implements FilterInterface
 {
 
     use InstanceConfigTrait;
@@ -90,9 +90,19 @@ abstract class Base
     /**
      * Get the manager.
      *
-     * @return \Search\Manager
+     * @deprecated Use getManager()
      */
     public function manager()
+    {
+        return $this->getManager();
+    }
+
+    /**
+     * Get the manager.
+     *
+     * @return \Search\Manager
+     */
+    public function getManager()
     {
         return $this->_manager;
     }
--- a/src/Model/Filter/Base.php
+++ b/src/Model/Filter/Base.php
@@ -90,19 +90,9 @@ abstract class Base implements FilterInterface
     /**
      * Get the manager.
      *
-     * @deprecated Use getManager()
-     */
-    public function manager()
-    {
-        return $this->getManager();
-    }
-
-    /**
-     * Get the manager.
-     *
      * @return \Search\Manager
      */
-    public function getManager()
+    public function manager()
     {
         return $this->_manager;
     }
--- a/src/Model/Filter/FilterInterface.php
+++ b/src/Model/Filter/FilterInterface.php
@@ -74,16 +74,6 @@ interface FilterInterface
     public function getArgs();
 
     /**
-     * Get / Set the validation rules.
-     *
-     * @param array|null $value Value.
-     * @return array|null
-     * @codeCoverageIgnore
-     * @internal
-     */
-    public function validate(array $value = null);
-
-    /**
      * Sets the query object.
      *
      * @param \Cake\Datasource\QueryInterface $value Value.
--- a/src/Model/Filter/FilterInterface.php
+++ b/src/Model/Filter/FilterInterface.php
@@ -84,15 +84,6 @@ interface FilterInterface
     public function validate(array $value = null);
 
     /**
-     * Valid method.
-     *
-     * @return bool
-     * @codeCoverageIgnore
-     * @internal
-     */
-    public function valid();
-
-    /**
      * Sets the query object.
      *
      * @param \Cake\Datasource\QueryInterface $value Value.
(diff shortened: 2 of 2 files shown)


src/Model/Filter/Base.php
```diff
@@ -90,9 +90,19 @@ public function __construct($name, Manager $manager, array $config = [])
...
+    {
+        return $this->getManager();
+    }
+
+    /**
+     * Get the manager.
+     *
+     * @return \Search\Manager
+     */
+    public function getManager()
```
ADmad: There's no setter for the manager so adding "get" prefix to the getter method is unnecessary. Let's avoid unnecessary method name changes.

src/Model/Filter/FilterInterface.php
```diff
@@ -0,0 +1,117 @@
...
+    public function validate(array $value = null);
+
+    /**
+     * Valid method.
+     *
+     * @return bool
+     * @codeCoverageIgnore
+     * @internal
+     */
+    public function valid();
```
ADmad: Having an internal method added to interface is weird.

burzum: Why is it public?

src/Model/Filter/Base.php
burzum: Well, I disagree with that, also on the framework level. It's weird.

ADmad:
What's the motivation for adding this interface besides technical correctness?

src/Model/Filter/FilterInterface.php
ADmad: IDK :)  I never noticed this before.

burzum:
A follow up PR that will add `FilterCollection`.

I want to get the huge declaration of a filter collection out of my table objects and would like to be able to do something like:

```php
$this->addBehavior('Search.Search', 'defaultFilterCollection' => new MyCollection());
```

src/Model/Filter/FilterInterface.php
```diff
@@ -0,0 +1,108 @@
...
+
+    /**
+     * Get / Set the validation rules.
+     *
+     * @param array|null $value Value.
+     * @return array|null
+     * @codeCoverageIgnore
+     * @internal
+     */
+    public function validate(array $value = null);
```
ADmad: valid() and validate() methods have no test coverage. It seems they are currently unused. Probably the plan was to add validation support for the search fields but it was never implemented.

burzum: We should remove it from the interface and code and add a second one `ValidateableFilterInterface` in the future?

burzum:
@ADmad I've moved the buildFilter() to a locator class, do you want a separate PR for that?

src/Model/Filter/FilterInterface.php
ADmad: Thanks sounds reasonable.

dereuromark:
At some point we also want to try raising phpstan from 1 to 3 etc, since only then it does a really thorough job in finding inter-class issues.

ADmad:
@dereuromark I just with phpstan level 5 locally and the only "real" issue is this:

```
 ------ ------------------------------------------------------------------------------
  Line   src/Model/Filter/Like.php
 ------ ------------------------------------------------------------------------------
  150    Call to an undefined method Cake\Datasource\QueryInterface::getConnection().
 ------ ------------------------------------------------------------------------------
```

https://github.com/FriendsOfCake/search/blob/8e576aa05349bd2e7cda00c0360da9f0ca4fb567/src/Model/Filter/Like.php#L150

burzum:
@ADmad how do you want to fix this? I would like to get these things merged some time soon instead of adding more and more and having more discussions.

dereuromark:
We can clean that up in a follow up PR.

ADmad:
@burzum Please have some patience and don't try to rush the process. There will be discussions for non trivial changes. I (and @dereuromark) have being giving feedback/comments regularly.

src/Model/Filter/FilterInterface.php
```diff
@@ -0,0 +1,98 @@
...
+ * Filter Interface
+ */
+interface FilterInterface
+{
+    /**
+     * Get the database field name.
+     *
+     * @return string|array
+     */
+    public function field();
```
ADmad: `field()` can also be dropped and deprecated in `Base`. It's better to just use `fields()` always which has a consistent return type.

ADmad:
I think this is good to go now. There are docblocks which have `\Search\Model\Filter\Base[]` which should be changed to `\Search\Model\Filter\FilterInterface[]` now.

src/Model/Filter/FilterInterface.php
```diff
@@ -0,0 +1,91 @@
...
+     */
+    public function getArgs();
+
+    /**
+     * Sets the query object.
+     *
+     * @param \Cake\Datasource\QueryInterface $value Value.
+     * @return void
+     */
+    public function setQuery(QueryInterface $value);
```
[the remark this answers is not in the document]
ADmad: > Query will always be implementation of QueryInterface.

Yes it will be but the issue is `ORM\Query` has additional methods which are not part of `QueryInterface`. That's the reason we can use the `QueryInterface` typehitn. Now we also want to allow this plugin to be usable with any data source which implements `RepositoryInterface` and not just `ORM\Table` as done through #203. The query objects for such repositories will only implement `Datasource\QueryInterface` (for e.g. muffin/webservice plugin). Hence the need for allowing `\Cake\Datasource\QueryInterface` or `\Cake\ORM\Query` instance.

ADmad:
Merging this to ensure @burzum doesn't become unhappy :smile:. I'll make a separate PR to take care of the typehint changes.

dereuromark:
They might yeah, but they don't have to.
It is just a bit cleaner from a contract perspective. It forces the dev/plugin to be make a more stable release and think better about the changes done per refactor.
Adding new public methods into the Base class yields the same problems, just a bit more dirty, btw.

PR #222, opened 2018-05-28
Moving filter loading into a locator

burzum:


src/Manager.php
```diff
@@ -231,18 +233,7 @@ public function custom($name, array $config = [])
...
-            $class = $options['className'];
-            unset($options['className']);
-        }
-        $className = App::className($class, 'Model\Filter');
-        if (!$className) {
-            throw new InvalidArgumentException(sprintf('Search filter "%s" was not found.', $class));
-        }
-
-        return new $className($name, $this, $options);
+        return (new FilterLocator($this))->locate($name, $filter, $options);
```
ADmad: Why build a new locator instance for each filter?  A single locator instance should suffice.

Unless the locator class is going to have more methods having a class with single method seems a bit of over engineering. In PR #221 you mentioned adding a `FilterCollection` (I like that idea). So couldn't that class handle locating/building instance too? Similar to what `WidgetLocator` does in core.

dereuromark: Also, using a hardcoded and non-factorized `(new FilterLocator($this))` seems like an anti-pattern to me.
This should at least then be a protected method `locator()` etc so you can exchange this if needed for your own locator when extending the manager etc.

src/Manager.php
```diff
@@ -40,6 +50,19 @@ class Manager
     public function __construct(RepositoryInterface $repository)
     {
         $this->_repository = $repository;
+        $this->_filterLocator = new FilterLocator($this);
+    }
+
+    /**
```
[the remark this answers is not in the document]
burzum: I don't get it why it complains here?

src/Manager.php
burzum: @ADmad I guess you want to drag that for BC reasons along anyway, so can't we just stick with the locator and use it in the manager and in the collection when it's done?

src/Manager.php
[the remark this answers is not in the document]
dereuromark: Because you are missing the param here.

src/Model/Filter/FilterLocator.php
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     * Loads a search filter.
+     *
+     * @param string $name Name of the field
+     * @param string $filter Filter name
+     * @param array $options Filter options.
+     * @return \Search\Model\Filter\Base
+     * @throws \InvalidArgumentException When no filter was found.
+     */
+    public function locate($name, $filter, array $options = [])
```
ADmad: It would better to use method name `get()` akin to core's `TableLocator::get()` and `WidgetLocator::get()`.

src/Manager.php
```diff
@@ -231,18 +255,7 @@ public function custom($name, array $config = [])
...
-            $class = $options['className'];
-            unset($options['className']);
-        }
-        $className = App::className($class, 'Model\Filter');
-        if (!$className) {
-            throw new InvalidArgumentException(sprintf('Search filter "%s" was not found.', $class));
-        }
-
-        return new $className($name, $this, $options);
+        return $this->_filterLocator->locate($name, $filter, $options);
```
ADmad: Method name needs to be updated to `get()` now :)

src/Manager.php
```diff
@@ -40,6 +50,20 @@ class Manager
     public function __construct(RepositoryInterface $repository)
     {
         $this->_repository = $repository;
+        $this->_filterLocator = new FilterLocator($this);
```
dereuromark: is it possible to extract this into a class var (static or otherwise) similar to what is done in core?
This would allow at least basic extension here - and sounds cleaner in terms of SOLID (especially since we dont use factory pattern here at all).

ADmad: Can also make the locator injectable through contructor and create an instance only if relevant argument is `null`.

dereuromark: Sure, both is fine.

ADmad:
Possible to improve code coverage to keep codecov happy? No biggie if not.

burzum:
@ADmad I'm now busy with other things. Also I'll probably never understand this silly Coveralls. If I click on any of the red blocks I get just "File not found in report" and it doesn't show me what exactly is not covered. Pretty useless. So I'll have to generate the report locally.

dereuromark:
Yeah codecov (not coveralls) is a pretty silly tool here in terms of file display.
It never worked for me either. We might have some setting here wrong, though.

ADmad:
Never mind then. The drop shown isn't significant.

src/Manager.php
burzum: @dereuromark are you going to make this change?

burzum:
Maybe we should look for another code coverage service that is actually useful and presents the results in a meaningful way?

dereuromark:
Any proposals? I know that coveralls wasnt much better either.

burzum:
@ADmad what other changes do you need before you would merge it?

dereuromark:
Pretty sure https://github.com/spryker/code-sniffer/blob/master/Spryker/ruleset.xml#L119 includes and runs this normally.

src/Manager.php
dereuromark: I guess with the existing setFilterLocator() setter this is probably no big deal here now.

src/Model/Filter/FilterLocator.php
```diff
@@ -0,0 +1,56 @@
...
+    public function get($name, $filter, array $options = [])
+    {
+        if (empty($options['className'])) {
+            $class = Inflector::classify($filter);
+        } else {
+            $class = $options['className'];
+            unset($options['className']);
+        }
+
+        $className = App::className($class, 'Model\Filter');
```
dereuromark: Isnt it `Model/Filter`? Like Model/Behavior etc?

tests/TestCase/Model/Filter/FilterLocatorTest.php
```diff
@@ -0,0 +1,29 @@
...
+     * @return void
+     */
+    public function testLoadFilter()
+    {
+        $articles = TableRegistry::get('Articles');
+        $manager = new Manager($articles);
+        $locator = new FilterLocator($manager);
+
+        $result = $locator->get('test', 'Search.Value');
+        $this->assertInstanceOf('\Search\Model\Filter\Value', $result);
```
dereuromark: No leading slash necessary, but ::class syntax would even be nicer and better for refactoring and IDE.

PR #223, opened 2018-05-29
Fix errors reported by phpstan at level 5.

ADmad:


phpstan.neon
```diff
@@ -0,0 +1,6 @@
+parameters:
+    autoload_files:
+        - tests/bootstrap.php
+    ignoreErrors:
+        - '#Call to an undefined method Cake\\Datasource\\RepositoryInterface::aliasField\(\)#'
+        - '#Call to an undefined method Cake\\ORM\\Table::searchManager\(\)#'
```
dereuromark: This should be possible to fix with mixin learning and proper annotation afaik.

phpstan.neon
```diff
@@ -0,0 +1,6 @@
+parameters:
+    autoload_files:
+        - tests/bootstrap.php
+    ignoreErrors:
+        - '#Call to an undefined method Cake\\Datasource\\RepositoryInterface::aliasField\(\)#'
```
dereuromark: This should also have an annotation that can be taught, right?

phpstan.neon
ADmad: Yes, but I didn't want to spend time adding a proper phpstan class reflection extension for now.

phpstan.neon
ADmad: Not that I know of. The code already has a `method_exists()` check before trying to use that method so there's no actual error. I don't know how to refactor the code to make phpstan aware of that (if it can be even made to understand).

dereuromark: We can also try to remove the ignors later of course.
Would have just been nice to not have to.

PR #225, opened 2018-05-31
Cleanup

ADmad:
@dereuromark While working on this I realized we have the same issue with use of typhint `Cake\Datasource\RepositoryInterface` as we have with `Cake\Datasource\QueryInterface`. The filters have code like `$repository->aliasField()` but the `aliasField()` method is not there in `RepositoryInterface`.

Adding `instanceof` checks makes phpstan happy and we don't need to ignore any errors or override variable types inline. So I think this is enough and we don't need to make any changes to the typehints. We can revisit them when we work on Cake 4 compatible version of the plugin, where hopefully the interfaces will be fixed up :slightly_smiling_face:.

### DIFF
--- a/src/Model/Filter/Base.php
+++ b/src/Model/Filter/Base.php
@@ -3,7 +3,6 @@ namespace Search\Model\Filter;
 
 use Cake\Core\InstanceConfigTrait;
 use Cake\Datasource\QueryInterface;
-use Cake\ORM\Table;
 use Search\Manager;
 
 /**
@@ -112,7 +111,7 @@ abstract class Base implements FilterInterface
         }
 
         $repository = $this->manager()->getRepository();
-        if (!$repository instanceof Table) {
+        if (!method_exists($repository, 'aliasField')) {
             return $field;
         }
 
--- a/src/Model/Filter/Base.php
+++ b/src/Model/Filter/Base.php
@@ -3,6 +3,7 @@ namespace Search\Model\Filter;
 
 use Cake\Core\InstanceConfigTrait;
 use Cake\Datasource\QueryInterface;
+use Cake\ORM\Table;
 use Search\Manager;
 
 /**
@@ -111,7 +112,7 @@ abstract class Base implements FilterInterface
         }
 
         $repository = $this->manager()->getRepository();
-        if (!method_exists($repository, 'aliasField')) {
+        if (!$repository instanceof Table) {
             return $field;
         }
 
--- a/src/Model/Filter/Base.php
+++ b/src/Model/Filter/Base.php
@@ -260,12 +260,12 @@ abstract class Base implements FilterInterface
     /**
      * Sets the query object.
      *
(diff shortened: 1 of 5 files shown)


src/Model/Filter/Base.php
```diff
@@ -111,7 +112,7 @@ public function field()
         }
 
         $repository = $this->manager()->getRepository();
-        if (!method_exists($repository, 'aliasField')) {
+        if (!$repository instanceof Table) {
```
dereuromark: Could this be a BC break for people implementing their own RepositoryInterface that does not extend Table?
Should this maybe be "or"?

dereuromark:
Sounds to me aliasField() has just been forgotten on the Interface as this is sure a very needed method on that level. After all setAlias() and getAlias() are very much part of it.
So it only makes sense to fix the interface (for BC add annotations here now) and keep the interface usage IMO?

tests/TestCase/Model/SearchTraitTest.php
```diff
@@ -63,8 +65,8 @@ public function testFinder()
 
         $query = $this->Articles->find('search', ['search' => $queryString]);
         $this->assertSame([
-            'Articles.foo' => 'a',
-            'Articles.public' => false,
+            'foo' => 'a',
```
dereuromark: This proves somewhat that there can be some BC break here in these changes (as so far it properly aliased).

ADmad: It's not BC break. The test was just wrong. `$this->Articles` is an instance of `Endpoint` class of `Webservice` plugin which doesn't have any `aliasField()` method.

ADmad:
> Sounds to me aliasField() has just been forgotten on the Interface as this is sure a very needed method on that level.

No it's not. Field aliasing is required only when doing joins on relational databases to avoid potential column name clashes. Custom repository classes for non relations datasource don't need it. Hence the `$repository instanceof Table` check on the line you commented on is good enough.

tests/TestCase/Model/SearchTraitTest.php
dereuromark: Interesting. OK

ADmad: Hmm no, I am wrong. `Endpoint` does have `aliasField()` method. So need to revert this :disappointed:

dereuromark: Yeah that was my gut feeling / expectation here.

dereuromark:
Based on the Endpoint class I think all kind of external non ORM solutions will most likely need a similar conflict resolution of aliasing to avoid naming conflicts.
So even if it is not part of this interface, maybe it should be soon part of some contracting interface then.

ADmad:
We will be able to remove the error ignore once https://github.com/phpstan/phpstan/issues/323 is taken care of.

Issue #224, opened 2018-05-29
Add ability to create filter collections.


ADmad:
Refs https://github.com/FriendsOfCake/search/pull/221#issuecomment-392539097 by @burzum.

A follow up PR that will add `FilterCollection`. I want to get the huge declaration of a filter collection out of my table objects.

So instead of having this:

```php
	public function configureSearch() : void {
		$this->searchManager()
			->add('id', 'Search.Value', [
				'field' => [$this->aliasField('id')],
			])
			->add('title', 'Search.Like', [
				'field' => [$this->aliasField('title')],
				'before' => true,
				'after' => true
			])
			->add('organizer', 'Search.Like', [
				'field' => [$this->aliasField('organizer')],
				'before' => true,
				'after' => true
			])
			->add('start_date', 'Search.Compare', [
				'field' => [$this->aliasField('start_date')],
				'operator' => '>='
			])
			->add('end_date', 'Search.Compare', [
				'field' => [$this->aliasField('end_date')],
				'operator' => '<='
			]);
	}
```

I want this:

```php
$this->addBehavior('Search.Search', 'defaultFilterCollection' => new MyCollection());
```

Or that:

```php
$this->searchManager()->setCollection('default', new MyCollection());
```

I also dislike the two-level array approach used in the manager. It's a collection of collections on the first level and then a collection of filters. This should be refactored.

ADmad:
Really like this idea. I think we can do even better  and avoid configuring the behavior/manager before hand.

It should not be difficult to have an implementation such that if the collection name specified in `find()` options does not exist then the search `Manager` looks for matching collection class (in `App\Model\Filter\) and instantiates it.

burzum:
@ADmad get my other PRs merged and I'll finish this! Hurry, hurry! 😛

burzum:
PR is open https://github.com/FriendsOfCake/search/pull/228

PR #229, opened 2018-06-19
Lazy load filter collection

ADmad:
This achieves the lazy loading of filter collection that I had stated in https://github.com/FriendsOfCake/search/issues/224#issuecomment-392764849.

- You can now have filter collection in your app e.g. `App\Model\Filter\MySweetCollection.php`.
  Use the `FilterCollection::initialize()` method to setup filters using `FilterCollection::add()`.
- Add search behavior to your table, no need to use search manager to setup filters.
- Specify the collection name in find call e.g. 
  `find('search', ['collection' => 'my_sweet', 'search' => $queryParams])`
  and viola!.

P.S. Search manager no longer loads filters directly anymore, nor does it maintain array of filters arrays. It only has array of collections now and filter loading is always delegated to the collection.

### DIFF
--- a/src/Manager.php
+++ b/src/Manager.php
@@ -1,7 +1,9 @@
 <?php
 namespace Search;
 
+use Cake\Core\App;
 use Cake\Datasource\RepositoryInterface;
+use Cake\Utility\Inflector;
 use InvalidArgumentException;
 use Search\Model\Filter\FilterCollection;
 use Search\Model\Filter\FilterCollectionInterface;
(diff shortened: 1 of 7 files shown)


src/Manager.php
```diff
@@ -17,13 +24,11 @@ class Manager
     protected $_repository;
 
     /**
-     * Filter collection and their filters
+     * Filter collections
      *
-     * @var array
+     * @var Search\Model\Filter\FilterCollectionInterface[] Filter collections list.
```
dereuromark: missing \

dereuromark:
Looks like only some docs are missing.
Maybe this would be a good time to start moving things into a /docs/README.md etc? And link from the main one - that main one should maybe only contain relevant install parts, usage could be better fit into a subdoc IMO.

ADmad:
With the end user feature enhancement now achieved I again question the need for filter locator.

It's seems like over engineering and an unnecessary layer of abstraction to have an interface and class, for what could instead be directly achieved by having a method with 5-6 lines of code in collection class.

PR #231, opened 2018-07-13
Add FilterMethodsTrait.

ADmad:
This allows using the convenience methods for filters on both `Manager` and `FilterCollection` instances.

### DIFF
--- a/src/Model/Filter/FilterCollection.php
+++ b/src/Model/Filter/FilterCollection.php
@@ -2,13 +2,14 @@
 namespace Search\Model\Filter;
 
 use ArrayIterator;
-use Search\Model\Filter\FilterLocatorInterface;
 
 /**
  * FilterCollection
  */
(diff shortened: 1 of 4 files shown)


src/Model/Filter/FilterCollection.php
```diff
@@ -2,13 +2,14 @@
...
 
 use ArrayIterator;
-use Search\Model\Filter\FilterLocatorInterface;
 
 /**
  * FilterCollection
  */
 class FilterCollection implements FilterCollectionInterface
 {
+    use FilterMethodsTrait;
```
dereuromark: Why does the filter collection need those?

ADmad: Because with my lazy loading filters support added in previous PR, one directly setups filters in collection class itself inside it's `initialize()` instead of using search manager.

PR #235, opened 2018-07-16
Added a basic contents table links

davidyell:
The readme is kinda big, so this pull request aims to mitigate that by adding a contents 'table' of links, to get to specific parts quicker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,9 +35,9 @@ Plugin::load('Search');
 
 ## Contents
 * [Usage](#usage)
-** [Table behaviour configuration](#table-class)
-** [Controller method configuration](#controller-class)
-** [Component](#component)
+ * [Table behaviour configuration](#table-class)
+ * [Controller method configuration](#controller-class)
+ * [Component](#component)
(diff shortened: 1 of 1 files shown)


dereuromark:
:-1: 
I actually just proposed a few days ago to keep readme lean and move the docs mainly to a subdir /docs as recommended in general.
Then this becomes not necessary anymore, or could be part of the /docs README one.

davidyell:
I don't see why we can't merge this then in the meantime. As when the pull request for the next update comes through, we can overwrite this change.

dereuromark:
Why not moving it already now?
Keeping this just invites the readme to get further exploded by too much content.

davidyell:
Sure, go for it 👍

dereuromark:
Do you want to move them to /docs/README.md ?
Keeping only the outline and basics in the readme then?

davidyell:
I don't see any value in such a small change. If we are going to use a `/docs` folder, there should be documentation for each logical section of the plugin, with links in the main repo readme.

PR #237, opened 2018-08-13
Dev

ADmad:
The public API remains backward compatible so I think just minor version bump should suffice when making a new release.


ADmad:
Should we add a new method `Manager::getFillterCollection()` which return `FilterCollectionInterface` and deprecate the `Manager::getFilters()` method?

ADmad:
phpstan is unhappy since `getConfig()` is called on filter instance but `FilterInterface` doesn't have that method :unamused:

dereuromark:
annotate for now till next major?

ADmad:
Can't probably just add the method to `FilterInterface`. All existing filter would be extending `Base` filter which uses `InstanceConfigTrait` and hence would have `getConfig()` implemented.

ADmad:
This shows having an interface as well as base class doesn't add much value, just adds pain :)

dereuromark:
In this specific case, having an abstract class to work with and typehint against, an interface contracting approach is indeed not adding too much value.

ADmad:
I'll remove the interface. No one is going to directly implement the interface instead of extending the `Base` filter.

ADmad:
To facilitate lazy coding I have updated `SearchBehavior` to conventionally use a collection with name matching the table alias if it exists. So for example if you attach search behavior to `PostsTable` then by default `App\Model\Filter\PostsCollection` will be used if it exists.

README.md
```diff
@@ -106,68 +78,135 @@ class ExampleTable extends Table
...
+
+Apart from configuring filters through search mananger in your table class,
+you can also create them as separate collection classes. This helps in
+keeping your table's `initialize()` method uncluttered and the filters are lazy
+loaded only when actually used.
 
-        return $searchManager;
+```php
+// src/Model/Filter/Filter/PostsCollection.php
+namespace App\Model\Filter\PostsCollection;
```
dereuromark: Is the namespace not just ...\Filter ?

README.md
```diff
@@ -228,12 +248,12 @@ After including the trait you can use the searchManager by calling the
 
 ## Filtering your data
 Once you have completed all the setup you can now filter your data by passing
-query params in your index method. Using the `Article` example given above, you
-could filter your articles using the following.
+query params in your index method. Using theexample given above, you
```
dereuromark: space missing

src/Manager.php
```diff
@@ -61,203 +78,110 @@ public function getRepository()
...
-     * @return $this
-     */
-    public function like($name, array $config = [])
-    {
-        $this->add($name, 'Search.Like', $config);
+        $instance = new $className($this);
+        if (!$instance instanceof FilterCollectionInterface) {
+            throw new InvalidArgumentException(sprintf(
+                'The collection must be instance of FilterCollectionInterface. ' .
+                'Got instanceof "%s" instead',
```
dereuromark:
```suggestion
                'Got instance of "%s" instead',
```

src/Model/Filter/FilterCollection.php
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * FilterCollection
+ */
+class FilterCollection implements FilterCollectionInterface
+{
+    use FilterMethodsTrait;
+
+    /**
+     * @var array List of filter objects
```
dereuromark:
```suggestion
     * @var \Search\Model\Filter\Base[] List of filter objects
```

src/Model/Filter/Like.php
```diff
@@ -147,8 +148,10 @@ protected function _formatWildcards($value)
     protected function _setEscaper()
     {
         if ($this->getConfig('escaper') === null) {
-            /** @var \Cake\Database\Query $query */
```
dereuromark: Is the below code still autocompletable in IDE without this? $query->getConnection()->getDriver() etc?

src/Model/SearchTrait.php
```diff
@@ -129,7 +141,7 @@ protected function _extractParams($params, $filters)
      * ```
      *
      * @param array $params The parameters array to flatten.
-     * @param array $filters The array of filters with configuration
+     * @param \Search\Model\Filter\Base[] $filters The array of filters with configuration
      * @return array The flattened parameters array.
      */
     protected function _flattenParams($params, $filters)
```
dereuromark: Ideally we also typehint the array on such a protected methods to make sure they are of that type going in.

src/Manager.php
```diff
@@ -32,14 +37,26 @@ class Manager
      */
     protected $_collection = 'default';
```
dereuromark: There is a lot of magic string usage around `'default'`. Could we make this a class const to more easily reuse it in other places? Especially when we compare against it, typos here can easily lead to hard to find dev errors.

dereuromark:
Looks good to go.
Whats the release timeline/plan?

README.md
dereuromark: Thats my point: PostsCollection in the namespace then?
I expected `App\Model\Filter` only.

dereuromark: Also the file name above does not match the namespace.

ADmad:
> Whats the release timeline/plan?

This weekend :slightly_smiling_face: 

I am going to bump major version due to the significant amount of changes but most user should be update to new version without requiring any changes in their app code.

dereuromark:
If this is bc i prefer a minor.
All plugins using this as dependency otherwise need adjusting again.

ADmad:
From usage POV it is BC but the signature of `Manager::__construct()` has changed. Not sure if that's allowed in a minor by semver.

dereuromark:
I think most of this is really internal. So a minor (compared to the fallout of majoring) is acceptable IMO.
A major triggers really a lot of plugins to be incompatible and disallowing for it until adjusted via patch manually.

ADmad:
I want to make some technically non BC changes like making `Manager::getFilters()` return filter collection instance instead of array which would warrant a major version bump.

I don't think a new major would be a problem. Any app/plugin wanting to use the new filters collection feature would just have to update the plugin's version to `^5.0`, that's it.

dereuromark:
Fair enough.
Are there any other BC breaking things we want to do or clean up then?

dereuromark:
I was thinking about making some methods protected that should never have been public in the first place.
Please see

    Manager::loadFilter()

etc.

ADmad:
> Are there any other BC breaking things we want to do or clean up then?

Can't remember off top of my head but I'll go through the code and see if there's anything else which needs cleanup.

> Manager::loadFilter()

That method is already removed on this branch :)

dereuromark:
Yeah, I saw and updated my comment. Still think that method should not be exposed as public interface method, though :)

ADmad:
> Still think that method should not be exposed as public interface method, though :)

Hmm.. `FilterCollection::loadFilter()` is used by `FilterCollection` itself only so I guess making it protected and removing from `FilterCollectionInterface` is fine.

dereuromark:
Looks nice!

PR #240, opened 2018-10-08
Add multi value separator

dereuromark:
We need the multiValue func for a field:
```php
        $searchManager
            ->like('search', ['field' => ['name', 'issue'], 'multiValue' => true]);
```

Currently in each controller action for this we have to hack it a bit with this:
```php
        $searchQuery = $this->request->getQuery();
        if (!empty($searchQuery['search'])) {
            $searchQuery['search'] = explode('|', $searchQuery['search']);
        }
        $query = $this->ReleaseGroups->find('search', ['search' => $searchQuery]);
```

Not sure if this can be done differently, a bit cleaner.
But I think given the Base class could handle this easily, a multiValueSeparator would be a useful config addition:

```php
        $searchManager
            ->like('search', ['field' => ['name', 'issue'], 'multiValue' => true, 'multiValueSeparator' => '|']);
```

Allowing us to then DRY config this to be used in all such actions:
```
ps-1232|xyz-345|foobar-999
```

This should also work fine with wildcards etc.

If accepted I am happy to add tests here :)

### DIFF
--- a/src/Model/Filter/Base.php
+++ b/src/Model/Filter/Base.php
@@ -214,7 +214,7 @@ abstract class Base
             return $this->getConfig('multiValue') ? $value : null;
         }
 
-        if ($this->getConfig('multiValue') && $this->getConfig('multiValueSeparator')) {
+        if ($this->getConfig('multiValueSeparator')) {
             return explode($this->getConfig('multiValueSeparator'), $value);
         }
 
--- a/src/Model/Filter/Base.php
+++ b/src/Model/Filter/Base.php
@@ -196,12 +196,6 @@ abstract class Base
             return $value;
         }
 
-        if (!is_array($passedValue) ||
-            $this->getConfig('multiValue')
-        ) {
-            return $passedValue;
-        }
-
         return $value;
     }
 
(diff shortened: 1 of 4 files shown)


src/Model/Filter/Base.php
```diff
@@ -188,13 +190,34 @@ public function skip()
...
+            return $passedValue;
+        }
+
+        return $value;
+    }
+
+    /**
+     * @return string|array|null
+     */
+    public function passedValue()
```
ADmad: Having a new public method doesn't seem necessary. Can't think of a case where just calling `passedValue()` would be required without the additional stuff done in `value()`.

dereuromark: Oh, I wanted to make this protected actually. My bad.

dereuromark: I find it cleaner to wrap this into such a method instead of having too much nesting/logic inside value() itself.

dereuromark:
I wrapped it a bit cleaner to encapsulate the multiValue as protected method vs the fallback/default of the value() method itself and added some basic tests.

ADmad:
Yeah checking for `multiValue` does become redundant when you have `multiValueSeparator` set. Having the former set to `false` while the latter is set wouldn't make sense. So just
```php
if ($this->getConfig('multiValueSeparator')) {
```
is fine.

dereuromark:
Done, I also added a non-happy case test for the new config.

PR #241, opened 2018-11-12
Fix finder to allow setting options.

dereuromark:
I need an important small improvement/fix for finder config in search.

I was expecting this to work:
```php
        $searchManager
            ->finder('shop_type', ['finder' => 'bits', 'map' => ['bits' => 'shop_type'], 'type' => 'contain'])
            ...;
```
As the custom finder needs

    $this->find('bits', [..., 'type' => 'contain])

to work as expected.

Now at least this (own config to avoid noise) works:

```php
        $searchManager
            ->finder('shop_type', ['finder' => 'bits', 'map' => ['bits' => 'shop_type'], 'options' => ['type' => 'contain']])
```

updated docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -404,7 +404,7 @@ The following options are supported by all filters except `Callback` and `Finder
 #### `Finder`
 
 - `finder` (`string`) The [find type](https://book.cakephp.org/3.0/en/orm/retrieving-data-and-resultsets.html#custom-finder-methods) to use.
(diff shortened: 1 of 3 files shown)


davidyell:
No new test case?

dereuromark:
Not that trivial here.
Would need a test class probably to show the setup.

dereuromark:
Thank you!

Issue #206, opened 2018-01-21
Search helper for resetUrl?


dereuromark:
Would it make sense to add a small helper with a reset button functionality - or at least the url generation part of it?
Currently one would either end up with this multiple times, or need his own project level helper:

```php
if (!empty($_isSearch)) {
	echo $this->Html->link(__('Reset'), ['?' => array_intersect_key($this->request->query, array_flip(['sort', 'direction']))], ['class' => 'btn btn-default']);
}
```

The reset should reset the filter on paginations, but it shouldnt reset the order IMO.
See http://sandbox.dereuromark.de/sandbox/search-examples/table?search=A&sort=iso2&direction=asc

We could provide
```php
public function resetUrl(array $whitelist = []) 
{
    $whitelist += $this->config('whitelist');

    return ['?' => array_intersect_key($this->request->query, array_flip($whitelist));
}
```

ADmad:
Your proposed `resetUrl()` is pretty generic and has no code related to the search filters. If we were to provide such a method it would make lot more sense to have one which removes the applied filters from query string. That could possibly be achieved by making `SearchBehavior::$_isSearch` (or add a new property) which maintains the list of names of filters applied.

dereuromark:
True, currently it would be very generic and "fit" by default only to a generic use case of core Paginator.

Some `isSearch` could indeed track the fields as array (fieldName, fieldName2, ...) which would then make it possible to extract those from the query string generated in this new SearchHelper.

PR #242, opened 2018-11-13
Add basic search reset button.

dereuromark:
Implements https://github.com/FriendsOfCake/search/issues/206

Before:
```php
    if (!empty($_isSearch)) {
        echo $this->Html->link(__('Reset'), 
            ['?' => array_intersect_key($this->request->query, array_flip(['sort', 'direction']))], 
            ['class' => 'btn btn-default']
        );
    }
```

Now:
```php
    if ($this->Search->isSearch()) {
        echo $this->Search->resetLink(__('Reset'), ['class' => 'btn btn-default']);
    }
```

If this looks good I can add some tests.


src/View/Helper/SearchHelper.php
```diff
@@ -0,0 +1,60 @@
...
+    {
+        return $this->Html->link($label, $this->resetUrlArray(), $options);
+    }
+
+    /**
+     * Returns the cleaned URL.
+     *
+     * @return array URL with cleaned Query string.
+     */
+    public function resetUrlArray()
```
ADmad: Method name can be just `resetUrl()`.

src/View/Helper/SearchHelper.php
```diff
@@ -0,0 +1,60 @@
...
+    {
+        $query = $this->request->getQuery();
+
+        $searchParams = $this->_View->get('_searchParams');
+        foreach ($searchParams as $searchParam => $value) {
```
ADmad: Could use `array_diff_key()` instead of the `foreach` loop?

dereuromark:
resetLink() test case is pending resolution of https://github.com/cakephp/cakephp/issues/12727

src/View/Helper/SearchHelper.php
```diff
@@ -36,21 +65,24 @@ public function isSearch()
...
         $query = $this->request->getQuery();
 
-        $searchParams = $this->_View->get('_searchParams');
-        foreach ($searchParams as $searchParam => $value) {
-            unset($query[$searchParam]);
+        $searchParams = (array)$this->_View->get('_searchParams');
+        $query = array_diff_key($query, $searchParams);
+
+        $additionalBlacklist = (array)$this->getConfig('additionalBlacklist');
+        foreach ($additionalBlacklist as $param) {
```
ADmad: Again the looping can be avoided by using array_flip() and array_key_diff() ☺️

tests/TestCase/View/Helper/SearchHelperTest.php
```diff
@@ -0,0 +1,147 @@
...
+        $this->assertSame($expected, $result);
+    }
+
+    /**
+     * @return void
+     */
+    public function _testResetLink()
+    {
+        $result = $this->searchHelper->resetLink('Reset search filter', ['class' => 'button']);
+        dd($result);
```
ADmad: Needs to be cleaned up.

src/View/Helper/SearchHelper.php
dereuromark: Could, but I find this slightly more readable than other possible shenanigans. ^^

tests/TestCase/View/Helper/SearchHelperTest.php
dereuromark: See link to bug ticket above for clarification on how to fix this.

dereuromark:
Works like a charm for me here in the app.

src/View/Helper/SearchHelper.php
```diff
@@ -0,0 +1,92 @@
...
+     * Checks for pagination and if so, blacklist limit and page params.
+     *
+     * @param \Cake\View\View $View View
+     * @param array $config Config
+     */
+    public function __construct(View $View, array $config)
+    {
+        $this->request = $View->request;
+        if ($this->request->getParam('paging')) {
+            $this->_defaultConfig['additionalBlacklist'] = ['page'];
```
ADmad: `$this->_defaultConfig['additionalBlacklist'][] = 'page'` would be a tad better, just in case someone wants to extend the helper and override `$_defaultConfig`.

dereuromark: Interesting edge case, sure thing.

ADmad:
Anything more to be done here?

dereuromark:
I think good to go.

PR #248, opened 2019-03-12
Add Exists filter. 

dereuromark:
Also fixed bad inflection and wrong filter result return value.

We often have to filter by content of rows (filled or not yet filled), and have a dropdown with
```php
// For a specific row field
'' => ' - show all - '
'1' => ' content set '
'0' => ' empty '
```
This Exists one takes the heavy load of a manual custom filter away as this easily does this based on the correct empty comparison for the ORM.
One improvement could be auto-detect of nullValue (schema not null => empty string), but maybe thats a bit too much overhead then.
So I settled for this compromise.

A new minor is OK?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -15,5 +15,4 @@ charset = utf-8
 end_of_line = crlf
 
 [*.yml]
-indent_style = space
 indent_size = 2
--- a/src/Model/Filter/FilterCollection.php
+++ b/src/Model/Filter/FilterCollection.php
@@ -74,7 +74,7 @@ class FilterCollection implements FilterCollectionInterface
     protected function _loadFilter($name, $filter, array $options = [])
     {
         if (empty($options['className'])) {
-            $class = Inflector::classify($filter);
+            $class = $filter;
         } else {
             $class = $options['className'];
             unset($options['className']);
--- a/docs/README.md
+++ b/docs/README.md
@@ -283,6 +283,7 @@ easily create the search results you need. Use:
 - `Boolean` to limit results by truthy (by default: 1, true, '1', 'true', 'yes', 'on')
   and falsy (by default: 0, false, '0', 'false', 'no', 'off') values which are
   passed down to the ORM as true/1 or false/0 or ignored when being neither truthy or falsy.
+- `Exists` to produce results for existing (non-empty) column content.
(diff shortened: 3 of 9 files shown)


.editorconfig
```diff
@@ -15,5 +15,4 @@ charset = utf-8
 end_of_line = crlf
 
 [*.yml]
-indent_style = space
```
ADmad: Why this change? YAML files are space indented.

src/Model/Filter/FilterCollection.php
```diff
@@ -74,7 +74,7 @@ public function add($name, $filter, array $options = [])
     protected function _loadFilter($name, $filter, array $options = [])
     {
         if (empty($options['className'])) {
-            $class = Inflector::classify($filter);
```
ADmad: Why this change?

ADmad:
> Also fixed bad inflection and wrong filter result return value.

Please make a separate PR for these so that a new bugfix release can be made. Then we can make a new minor with the new filter.

.editorconfig
dereuromark: Your ide will tell u: not necessary here.

src/Model/Filter/FilterCollection.php
dereuromark: Revert and u see: bad code killing the name resolvement for things that are plural or appear as such

dereuromark:
IMO those can and should be shipped together.

ADmad:
The bug fixes and new filter are unrelated. The bugfixes should first go in a bugfix release. After that a new minor with `Exists` filter.

src/Model/Filter/FilterCollection.php
ADmad: That's what tests are for :wink:.

dereuromark:
I wanted keep the process lean here. A bugfix can always go into a minor just fine.
Feel free to extract them, but people could even have relied on certain wrong behavior .
Shipping this in this minor then even makes more sense than patching it out.

ADmad:
You won't have to come up with justifications if you simply stop clubbing unrelated changes in same PR.

.editorconfig
ADmad: The `editorconfig` file is meant to tell the formatting style to IDE.

ADmad: The proper justification for removing this is it's redundant since we already have `indent_style = space` under `[*]`.

PR #250, opened 2019-05-16
Add beforeProcess callback to all filters

### LINKED ISSUE #249: Allow modify query before applying filter
I have to search by an associated column, so I used a like, but sometimes the query doesn't have the join to search, resulting in a bad sql.
I propose add this callable to keep with the filter instead of manipulating the query in every place that calls the search finder, the callable should be called before calling process()

```php
// In ArticlesTable
$manager->like('author', [
    'field' => [
        'Authors.name',
    ],
    'before' => true,
    'after' => true,
    'wildcardAny' => '*',
    'wildcardOne' => '?',
    // this will be added, maybe called beforeProccess
    'beforeFilter' => function ($query, $filter) {
        $query->leftJoinWith('Authors');
    }
]);
```

raul338:
Closes #249
Allows manipulation of the query as "setup" for the filters.

Should I modify `testProcessFilters` to expect beforeProcess or How should I test in SearchBehavior?

### DIFF
--- a/src/Model/Filter/Base.php
+++ b/src/Model/Filter/Base.php
@@ -68,6 +68,7 @@ abstract class Base
             'multiValue' => false,
             'multiValueSeparator' => null,
             'flatten' => true,
+            'beforeProcess' => null,
         ];
         $config += $defaults;
         $this->setConfig($config);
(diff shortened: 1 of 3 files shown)


ADmad:
Thanks for the pull request but closing in favor of #251 :slightly_smiling_face:.

PR #251, opened 2019-05-16
Processor

ADmad:
While looking into #249, the fact that the guts for filter processing were tucked away in a trait irked me. So I moved the relevant code into a separate class.

Making the filters invokeable is an elegant solution for running the "beforeProcess" callback and also not having to call various other methods of the filter explicitly before running `process()`.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -329,7 +329,9 @@ The following options are supported by all filters except `Callback` and `Finder
 - `beforeProcess` (`callable`, defaults to `null`) A callable which can be used
   to modify the query before the main `process()` method of filter is run.
   It receives `$query` and `$args` as arguments. You can use the callback for e.g.
-  to setup joins or contains on the query.
+  to setup joins or contains on the query. If the callback returns `false` then
+  processing of the filter will be skipped. If it returns `array` it will be used
+  as filter arguments.
 
 #### `Boolean`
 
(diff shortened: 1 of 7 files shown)


ADmad:
I am wondering if callbacks can be made more powerful by allowing them to return `false` to skip running `process()` or return an array to be used as `$args`.

dereuromark:
It could make sense as right now it seems to be an all or nothing kind of callback, making it more difficult to use it fine-tuned based on the different possible input coming in.

Sounds like both ideas make sense to allow.

ADmad:
I'll make a new minor release soon.

docs/README.md
```diff
@@ -326,11 +326,18 @@ The following options are supported by all filters except `Callback` and `Finder
 - `defaultValue` (`mixed`, defaults to `null`) The default value that is being
   used in case the value passed for the corresponding field is invalid or missing.
 
+- `beforeProcess` (`callable`, defaults to `null`) A callable which can be used
+  to modify the query before the main `process()` method of filter is run.
+  It receives `$query` and `$args` as arguments. You can use the callback for e.g.
```
dereuromark: We might need an example of the structure of such a callback. An example on how to use it for default case should suffice (void/null return).

ADmad: Done in https://github.com/FriendsOfCake/search/commit/8731c04c637198ae0fb37e943f617229a8997931

PR #254, opened 2019-06-03
Allow persisting nested query params (e.g. scoped pagination)

swiffer:
Expands an array that was previously flattened with Hash::flatten():


ADmad:
Can you please provide a test case for this.

swiffer:
@ADmad done ;)

PR #256, opened 2019-09-09
Add PHP74 to matrix, fix travis run

dereuromark:


dereuromark:
PHP74 is currently allowed to fail until we have a good solution.

Note: one possible solution is shown here: https://github.com/dereuromark/cakephp-shim/pull/61/files

PR #257, opened 2019-10-02
Set default label for reset link.

ADmad:


ADmad:
The build failure are due to CS failures. I will fix them separately.

PR #270, opened 2020-03-16
Fix errors reported by psalm

ADmad:


dereuromark:
Property Search\Manager::$_collectionClass has no typehint specified.

ADmad:
> Property Search\Manager::$_collectionClass has no typehint specified.

It's does, just phpstan is not smart enough to understand it :)

dereuromark:
I figured so.

PR #271, opened 2020-03-16
Removed deprecated method Base::field().

ADmad:
Rename filter option "field" to "fields".

### DIFF
--- a/src/Model/Filter/Base.php
+++ b/src/Model/Filter/Base.php
@@ -59,8 +59,8 @@ abstract class Base
     {
         $this->_manager = $manager;
 
-        $defaults = [
-            'field' => $name,
+        $config += [
+            'fields' => $name,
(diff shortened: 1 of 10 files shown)


src/Model/Filter/Base.php
```diff
@@ -72,12 +72,18 @@ public function __construct(string $name, Manager $manager, array $config = [])
...
+        if (isset($config['field'])) {
+            throw new \InvalidArgumentException(
+                'The `field` option has been renamed to `fields`.'
+            );
+        }
+        $config['fields'] = (array)$config['fields'];
         $this->setConfig($config);
 
         if (
-            (empty($config['field']) && $config['field'] !== '0') ||
```
dereuromark: Is this safe to change? As the check is specifically around the falsey values in php.

PR #275, opened 2020-03-22
Remove unused methods.

### LINKED ISSUE #272: Drop `Base::valid()/validate()` in v6
These methods are remnant of an unimplemented feature. There's no point carrying them along for new major version. They can be also re-added in future if/when the related feature is actually implemented.

ADmad:
Closes #272


dereuromark:
If you look into the commit of 2014 that introduced this, there are more pieces to this.
E.g.

    'validate' => [],

in Base class etc.

Issue #265, opened 2020-02-24
Search is throwing AuthorizationRequiredException


davidspeijer:
I'm experiencing the same issues as in #258.

my function in the controller:
```
public function index() {
    $this->paginate = [
        'contain' => [
            'Companies',
            'ShippingBarcodes' => ['sort' => ['ShippingBarcodes.id' => 'DESC']]
        ],
        'order' => ['Orders.id' => 'DESC']
    ];
    
    $orders = $this->Orders->find('search', ['search' => $this->request->getQueryParams()]);
    $this->Authorization->authorize($orders);
    $orders = $this->paginate($orders);

    $companies = $this->Orders->Companies->find('list')->where(['Companies.published' => true])->order(['Companies.name' => 'ASC']);
    $statuses = Order::status();

    $this->set(compact('orders','companies','statuses'));
}
```

This is loading the page as expected. The problems occurs once I try to filter the results. The POST results in a status 500: 

> The request to/orders/index did not apply any authorization checks.

Adding a skipAuth is not solving the problem:
```
public function index() {
    $this->Authorization->skipAuthorization();

    $this->paginate = [
        'contain' => [
            'Companies',
            'ShippingBarcodes' => ['sort' => ['ShippingBarcodes.id' => 'DESC']]
        ],
        'order' => ['Orders.id' => 'DESC']
    ];
    .........
}
```
As suggested in #258 I tried to catch the problem in beforeFilter();
```
public function beforeFilter(\Cake\Event\EventInterface $event)
{
    parent::beforeFilter($event);
    if ($this->request->getParam('action') == 'index' && $this->Authentication->getIdentity()->get('role') == 'admin') {
        $this->Authorization->skipAuthorization();
    }
}
```

This is "solving" the issue as a workaround.

However not as I expect it to work. As I assume that the `$this->Authorization->skipAuthorization()` or `$this->Authorization->authorize($orders)` in the index method should have applied the auth check even with the redirect?

dereuromark:
How is this related to this plugin?
It seems this is a side effect here and the actual issue happens from another plugin and code.

ADmad:
Skipping authorization for your method in `initialize()` or `beforeFilter()` is the correct way, not a work around.

Issue #258, opened 2019-10-03
Search triggers an error when used with Authorization plugin


alexvanniel:
When I perform a Search (submit the search form) while the Authorization plugin is active and loaded, I get the "The request to `xxxxxxx` did not apply any authorization checks.".
However, in the controller action the first thing that is handled is the authorization.

Does the Search (or maybe Search.Prg) do a redirect while searching and handling events and thus skipping the Authorization procedure? Or is there something I overlooked while configuring the Search plugin that is known to conflict with Authorization in combination with Search?

Ofcourse there is an option to skip authorization but that should not be necessary and far from secure.

ADmad:
The "R" in PRG is for redirect :slightly_smiling_face:. But it's to the same URL with additional query string. There's nothing else which would interfere with Authorization.

alexvanniel:
Yeah... I totally did not see that in the documentation even though it is there and written quite clearly even. I also figured that the R would stand for Redirect. Sorry for my brain fart.
It appears however that for some reason the Authorization middleware prevents the Prg plugin to redirect. The Authorization middleware will stop execution because it did not perform an authorize($model, $action) before the process was completed (which would be the actual redirect). So basically I am looking for a way to add authorize($model, $action) just before the Prg plugin redirects the user.

alexvanniel:
I think I found something that seems to work. When i add authorize() to the controllers beforeFilter method I prevent the middleware from failing authorization. I'd need to limit this to only those action the Prg plugin acts upon... Although it does feel a bit dirty.
I admit, this is not strictly a Search.Prg issue, but is triggered by it like the Security component can be triggered as per documentation if you do not specify the exception in there.

PR #282, opened 2020-03-25
Make search compatible with authorization component

davidspeijer:
The Search plugin does a redirect in its startup method. Because this event happens before controller action and aborts the request, the skipAuthorization() and authorize() calls in controller actions are never reached on the redirect response. cakephp/authorization#128

Moving the logic from the startup() to beforeRender() solves the issue.
#258 #265

### DIFF
--- a/src/Controller/Component/SearchComponent.php
+++ b/src/Controller/Component/SearchComponent.php
@@ -84,7 +84,7 @@ class SearchComponent extends Component
         if (!$this->_isSearchAction()) {
             return;
         }
-        
+
         if ($this->getController()->getRequest()->is('post') && $this->_isSearchAction()) {
             $url = $this->getController()->getRequest()->getPath();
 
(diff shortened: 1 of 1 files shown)


dereuromark:
I am not sure beforeRender() is a good idea for this.
At that point the query has been run for whatever was there before, creating quite some overhead on the system.
I think startup() etc should be kept.
The auth part just should happen before this

See https://github.com/dereuromark/executionorder/blob/master/RESULTS.md#redirecting
```
Controller::initialize
FooComponent::beforeFilter
Controller::beforeFilter
FooComponent::startup
Controller.action

Controller::redirect
FooComponent::beforeRedirect
```

=> `beforeFilter()` should be the part any auth hooks in.

davidspeijer:
Partly agreeing with you. 

Shall we continue to discuss at https://github.com/cakephp/authorization/issues/128 to also include (and potentially change) the authorization plugin?

ADmad:
Closing as there's nothing to be done in the search plugin.

PR #286, opened 2020-04-18
Use stable versions

dereuromark:
This seems to require first

    "muffin/webservice": "^3.0"

being released as stable.


ADmad:
Fixed in https://github.com/FriendsOfCake/search/commit/cb1566a7efcffe5b7fffe817e7bd0cbf0121ad3d.

PR #293, opened 2020-11-02
Add example for combining

dereuromark:
also fixed header

I found this out when trying to set a custom one but it still used the "inherited" one.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -91,7 +91,7 @@ Let's use the *backend*'s filters by doing:
     }
 ```
 
-## Filter collection classes
+#### Filter collection classes
 
 Apart from configuring filters through search mananger in your table class,
 you can also create them as separate collection classes. This helps in
@@ -150,6 +150,33 @@ You can also specify alternate collection class to use when making find call:
 ```
 
 The above will use `App\Model\Filter\PostsBackendCollection`.
+
+
+#### Collection class vs table config
+
+You can also set defaults in the Table class and inherit those for all searches.
+The added collection classes would then provide only custom ones per search.
+
+In your Table:
+```php
+    /**
+     * @return \Search\Manager
+     */
+    public function searchManager()
+    {
+        $searchManager = $this->behaviors()->Search->searchManager()
+            ->value('status');
+
+        return $searchManager;
+    }
+```
+In your Controller:
+```php
+    $this->Posts->addBehavior('Search.Search', [
+        'collectionClass' => PostsFilterCollection::class,
+    ]);
+```
+This would add additional filters on top of inherited `status` one.
 
 ### Search Component
 Add the `Search.Search` component with the necessary actions in your controller.


docs/README.md
```diff
@@ -151,6 +151,33 @@ You can also specify alternate collection class to use when making find call:
...
+        return $searchManager;
+    }
+```
+In your Controller:
+```php
+    $this->Posts->addBehavior('Search.Search', [
+        'collectionClass' => PostsFilterCollection::class,
+    ]);
+```
+This would add additional filters on top of inherited `status` one.
```
ADmad, line 179: Technically the ones you add in `searchManager()` are the additional ones that are added to the ones in `PostsFilterCollection`.

dereuromark: Interesting, as it still overwrote the ones in the filter for me.
I will have to test that once more.
Maybe it appends it and PHP just uses the last config in array found by key ;)

ADmad: Yeah using the same name will override, so if `PostsFilterCollection` already had a filter with name `status` it will be overwritten through `searchManager()`.

dereuromark: Which in the end means what I wrote though ;) Even if the technical detail might be a bit different.

dereuromark: I wonder if we should fix the code here to first use the generic ones, then add the custom ones (overwriting).

ADmad: When you specify `collectionClass` you are setting the "default" collection to use (which has preset filters). So if you are now using `searchManager()` too you are setting additional filters to that collection and if one of them has same name it would override the one from the collection class.

dereuromark: I dont know, I see it the other way around
I have the behavior attached without any config, so it reads the default Table one
Now I attach it to other actions with specific filters, now I expect those to take precedence.

So from no to some config explicitly set, I think my direction is the expected and intuitive one.

ADmad: > I have the behavior attached without any config

That's not what's done in the above example since `'collectionClass'` is specified when adding the behavior.

dereuromark: But it is.
The default config is in the Table::searchManager()
Now the collection gets attached on top.

That is the scenario I just mentioned (apart from the other locations where it is attached without any config rendering this case the default one).

ADmad: > The default config is in the Table::searchManager()
> Now the collection gets attached on top.

dereuromark: Yes, but with the wrong merge order if we go back to the beginning.
And I am saying that we should fix that.

In other words:
The default is what gets always applied (and possibly overwritten by extensions).

So the Table one is the default (or should be)
And custom filters are on top overwriting on conflict.

dereuromark: I dont agree with this
As you cannot modify the Table ones anymore.
So those are fixed and thus quite "default".
You can however mix in always a custom collection, so those would be the extension ones - or should be.

dereuromark: What you are saying renders the Table method useless, as it would be inflexible in the combination with any collectionClass manually defined (custom config).

So I propose to give it some value if used together.
If used in separation there is no issue anyway.

But if you dont see any value in this, I am also happy to close this one.

ADmad: There's no merging of filters/collections done anywhere. The use of `collectionClass` decides which collection class is instantiated and used for the name "default".

dereuromark: Feel free to rephrase this PR if needed, then we can probably close.
I guess I don't want to deep dive too much into this - even if there was some value for me.
I just then go either full class or full table for mine for now.

ADmad: > I just then go either full class or full table for mine for now.

That's the best approach. I personally would never specify `collectionClass` and use `searchManager()` too to modify the collection instance. That's why I don't want to document this and put ideas into people's minds :slightly_smiling_face:.

Just the use of `collectionClass` is already documented.

ADmad: Sounds fair, I'll think of a suitable note.

PR #294, opened 2020-11-02
Dont drop passed params for reset

dereuromark:
We forgot this at the time

When you have passed params, e.g. /controller/index/my-filter?query=xyz
Then the reset link would drop the "my-filter" and create 404s (when the passed params are expected/required).

This fixes it.


dereuromark:
Done

Issue #297, opened 2020-11-04
warning with bake in cake42


luizcmarin:
![image](https://user-images.githubusercontent.com/67489841/98117881-195fe580-1e89-11eb-8d08-a39a4ca3e2c6.png)

deprecation error when generating bake application.

dereuromark:
Can you open a PR with the suggested changes, please?

bear in mind that this can also create a barrier as a constraint update to that new minor is then required.
so all has its pros and cons
https://github.com/FriendsOfCake/search/blob/master/composer.json#L25 => will become ^4.1 then

luizcmarin:
Every VENDOR plugin must be updated to keep the entire product in perfect working condition.
I request the update of the plugin to adapt to the new changes in cakephp 4.2.
In particular, get the use of getTable for Table, as shown in the image and quotes
https://book.cakephp.org/4.next/en/appendices/4-2-migration-guide.html#orm

ADmad:
@luizcmarin Thanks for the reporting this. CakePHP 4.2 is no where close to a stable release, so there's no cause for alarm about a deprecation notice.

The issue will be addressed in due time, meanwhile you can continue trying 4.2 without any problem.

PR #298, opened 2020-11-04
Update "getTable" to "table"

luizcmarin:
I request the update of the plugin to adapt to the new changes in cakephp 4.2.
In particular, get the use of getTable for Table, as shown in 
https://book.cakephp.org/4.next/en/appendices/4-2-migration-guide.html#orm


dereuromark:
As per https://github.com/FriendsOfCake/search/issues/297#issuecomment-721839539 this will require a bump to even >= 4.2 on core
This sure isn't ideal for everyone, so I would say we hold off for now as long as possible.

dereuromark:
To avoid the bump we could add a method_exists() check
Since this action is only performed once on a normal request, this seems reasonable IMO.

ADmad:
> To avoid the bump we could add a method_exists() check

That's what I plan to do.

PR #299, opened 2020-11-04
Switch to Github Actions for CI.

ADmad:
TravisCI is ending free OSS support.


dereuromark:
Nice work :)

dereuromark:
Note: We should also switch the badge
https://github.com/spryker/decimal-object/pull/12 works nicely and links to the branch result only (no side effect of other branches for the display of master).

ADmad:
The badge URL I have used is for master only.

dereuromark:
Right. It looked different than the CI one.

PS: Do we want composer caching enabled?
https://github.com/actions/cache/blob/main/examples.md#php---composer

dereuromark:
This worked for me (using composer.json file to hash): https://github.com/dereuromark/cakephp-ide-helper/pull/220

ADmad:
I don't want to use GA cache until they add a way to be able to explicitly purge the cache.

This (and many other plugins) don't gets updates often enough so the lack of caching won't hurt much.

Issue #296, opened 2020-11-03
Bake filter collections?


dereuromark:
Do we want to add a bake task to easily generate filter collection classes?

With basic fields of scalar value added into as 
- string: like()
- key or int: value()

E.g.
```php
namespace App\Model\Filter;

use Search\Model\Filter\FilterCollection;

class ProfilesFilterCollection extends FilterCollection
{
    /**
     * @return void
     */
    public function initialize(): void
    {
        $this->like('title')
            ->value('country_id')
            ...;
    }
}
```

As it is easy to just remove a few lines, but it would be nice to have them generated.

We should also only add visible fields of the entity.

ADmad:
Sure, having such a bake task can't hurt.

dereuromark:
(Admin) prefixed ones would go into

> /Filter/Admin/...

Or how would you handle them?

PR #300, opened 2020-11-04
Bake task for filter collections

dereuromark:
I whipped sth up
Resolves https://github.com/FriendsOfCake/search/issues/296

I didn't check the visibility on the entity yet, if I should do that let me know
But this already works quite well for default cases IMO.

- Model handling if not matching the collection name
- Prefix handling

We could think about further improvements as follow up:
- No argument => ask [y/n] to bake each existing model of that namespace into the same name of collection

### DIFF
--- a/.stickler.yml
+++ b/.stickler.yml
@@ -5,6 +5,9 @@ linters:
     extensions: 'php'
     fixer: true
 
+files:
+  ignore: ['tests/comparisons/*']
+
 fixers:
   enable: true
   workflow: commit
--- a/src/Command/BakeFilterCollectionCommand.php
+++ b/src/Command/BakeFilterCollectionCommand.php
@@ -10,6 +10,7 @@ use Cake\Console\ConsoleIo;
 use Cake\Console\ConsoleOptionParser;
 use Cake\Core\Configure;
 use Cake\Database\Exception;
+use Cake\ORM\Table;
 
 /**
  * For generating filter collection classes.
@@ -49,6 +50,15 @@ class BakeFilterCollectionCommand extends BakeCommand
             'string',
             'char',
         ],
+    ];
+
+    /**
+     * @var string[]
+     */
+    protected $_ignoreFields = [
+        'lft',
+        'rght',
+        'password',
     ];
 
     /**
@@ -177,7 +187,7 @@ class BakeFilterCollectionCommand extends BakeCommand
 
         $fields = [];
         foreach ($columns as $column) {
-            if ($column === $table->getPrimaryKey() || $column === 'password') {
+            if ($this->shouldSkip($column, $table)) {
                 continue;
             }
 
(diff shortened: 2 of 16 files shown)


dereuromark:
This whole file system generation and comparison thing is a bit strange
For some reason it fails only for https://travis-ci.com/github/FriendsOfCake/search/jobs/428393497 still.

ADmad:
> This is prefer lowest, must be related to that.

That's the problem with using `prefer lowest`, you don't get the bugfixes required :)

ADmad:
My guess would be it's the twig view plugin for which a particular big bugfix is required.

dereuromark:
Otherwise all good? Or do you see any concerns so far?

dereuromark:
Tests are now fixed :)

src/Command/BakeFilterCollectionCommand.php
```diff
@@ -0,0 +1,225 @@
...
+            'tinyinteger',
+            'biginteger',
+            'smallinteger',
+            'uuid',
+            'binaryuuid',
+            'boolean',
+        ],
+        'like' => [
+            'string',
+            'char',
```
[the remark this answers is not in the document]
dereuromark: I would say no, as this contains often too much text and is not optimized for this kind of normal like searches.

tests/comparisons/Filter/testDefault.php
```diff
@@ -0,0 +1,19 @@
...
+use Search\Model\Filter\FilterCollection;
+
+class PostsFilterCollection extends FilterCollection
+{
+    /**
+     * @return void
+     */
+    public function initialize(): void
+    {
+        $this->value('author_id');
```
ADmad, line 15: Would be nice if instead of repeating `$this` on each line method chaining was used similar to https://github.com/cakephp/bake/blob/85be666e8fbcc4a67ff0c0c7a66221623793cd5a/tests/comparisons/Model/testBakeTableValidation.php#L51-L53

dereuromark: Yeah, for simplicity of the command, I decided not to do this.
Do you have an idea on how to easily achieve that?

For me, it seems enough as is, since it works both equally, and the lines don't increase.
So I am fine either way.

ADmad: Yes it's the same number of lines but avoid repeatation of `$this` would be nice. 

You can check bake's temple for tables and see how it generates the chaining. It should be just a matter of tweaking a loop but if it's too much trouble we can keep as it is.

ADmad:
CS job is still failing, wonder why stickler hasn't fixed it.

dereuromark:
The CS fails seemed unrelated, but I fixed them now as well
After rebase I expect this to be green now.

As for the $this chaining. I think this would need a helper method and some additional internal logic.
If you don't mind I would like to hold off on that for now as a future follow up (maybe together with the improvements above).

.stickler.yml
```diff
@@ -5,6 +5,9 @@ linters:
     extensions: 'php'
     fixer: true
 
+files:
+  ignore: ['tests/comparisons/*']
```
dereuromark, line 9: @markstory Any idea why this could be ignored in stickler still?
I copied this from https://github.com/cakephp/migrations/blob/master/.stickler.yml#L4-L5

markstory: Are you aiming to have these files *not* checked? That is what this config would do.

markstory: Unfortunately, FriendsOfCake is using the OAuth version of the integration so we won't see any logs here. If the GitHubApp integration was used we'd get better logs and annotations instead of comments.

ADmad: I have done Stickler's GitHubApp integration now for FOC.

ADmad: Stickler respects `phpcs.xml` right? So it would be better to add the ignores there @dereuromark.

markstory: @ADmad it does.

PR #303, opened 2020-11-09
Bake task for filter collections

dereuromark:
Replaces https://github.com/FriendsOfCake/search/pull/300 if is green.


dereuromark:
:tada:

PR #304, opened 2020-11-20
Allow not filter for value instead of long array list.

dereuromark:
The multiValue part is nice, but once you have too many values, the URL gets out of hand.

    ?functional_type=0,1,2,8,10,3,7,5,4,6,11,...........&...

For this I propose the "not" config instead.
Works fine for us locally.

    ?functional_type=!9&...

Does this sound good?

Fully BC.

Constraints: This is disabled to work together with multiValue as it is designed as alternative config.
If someone has ideas how to combine them, that could be a future PR maybe?

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -437,6 +437,11 @@ Set it to an empty string there to check via `=`/`!=` instead of `IS NULL`/`IS N
 - `mode` (`string`, defaults to `OR`) The conditional mode to use when matching
   against multiple fields. Valid values are `OR` and `AND`.
 
+- `not` (`string`, defaults to `null`) An alternative to `multiValue`, especially if you have
+ a lot of values. The filter accepts any string, but it should ideally be a single and unique char
+ as prefix. E.g. `!` for string values or `-` for numeric values.
+ If enabled, the filter will invert the expression for this value.
+
 #### `Finder`
 
 - `finder` (`string`) The [find type](https://book.cakephp.org/4/en/orm/retrieving-data-and-resultsets.html#custom-finder-methods) to use.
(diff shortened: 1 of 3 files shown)


docs/README.md
```diff
@@ -437,6 +437,11 @@ Set it to an empty string there to check via `=`/`!=` instead of `IS NULL`/`IS N
 - `mode` (`string`, defaults to `OR`) The conditional mode to use when matching
   against multiple fields. Valid values are `OR` and `AND`.
 
+- `not` (`string`, defaults to `null`) An alternative to `multiValue`, especially if you have
```
ADmad, line 440: More than a single character to denote negation doesn't seem necessary. So an option named `negationChar` would be more intuitive.

src/Model/Filter/Value.php
```diff
@@ -40,13 +40,13 @@ public function process(): bool
...
-        if ($this->getConfig('not')) {
+        if ($this->getConfig('negationChar')) {
             if ($this->getConfig('multiValue')) {
                 throw new Exception('Cannot use NOT functionality with multi value');
             }
 
-            if (strpos($value, $this->getConfig('not')) === 0) {
-                $value = mb_substr($value, strlen($this->getConfig('not')));
+            if (strpos($value, $this->getConfig('negationChar')) === 0) {
+                $value = mb_substr($value, strlen($this->getConfig('negationChar')));
```
ADmad, line 49: Should `mb_strlen` be used instead of `strlen` for consistency at least?

src/Model/Filter/Value.php
```diff
@@ -37,6 +39,18 @@ public function process(): bool
             return false;
         }
 
+        $isNot = false;
```
ADmad, line 42: `$isNegated` would be better.

PR #308, opened 2021-02-18
Fix return value for Boolean.

bartikowskiw:
The process method of the Boolean filter can return false even if the filter was actually applied. Which results i.E. in SearchHelper::isSearch() returning a wrong value. This should fix that issue.


ADmad:
Thanks for the PR. I fixed it slightly differently (with tests) in #309 and made a new release.

bartikowskiw:
Great. Thanks!

PR #313, opened 2021-10-01
fix SearchComponent load model error

smladeoye:
This update is to fix errors in the SearchComponent when it attempts to load a model that doesn't exist.

### DIFF
--- a/src/Controller/Component/SearchComponent.php
+++ b/src/Controller/Component/SearchComponent.php
@@ -99,8 +99,13 @@ class SearchComponent extends Component
 
         $controller = $this->getController();
         try {
+            // check to ensure modelclass is set before loading it
+            $model_class = $this->getConfig('modelClass');
+            if (!$model_class) {
+                return;
+            }
             /** @var \Cake\ORM\Table&\Search\Model\Behavior\SearchBehavior $model */
-            $model = $controller->loadModel($this->getConfig('modelClass'));
+            $model = $controller->loadModel($model_class);
         } catch (UnexpectedValueException $e) {
             return;
         }


src/Controller/Component/SearchComponent.php
```diff
@@ -99,8 +99,13 @@ public function beforeRender()
 
         $controller = $this->getController();
         try {
+            // check to ensure modelclass is set before loading it
+            $model_class = $this->getConfig('modelClass');
```
ADmad, line 103: As per Cake's coding standards the var name should be `$modelClass`.

src/Controller/Component/SearchComponent.php
```diff
@@ -99,8 +99,13 @@ public function beforeRender()
 
         $controller = $this->getController();
         try {
+            // check to ensure modelclass is set before loading it
```
ADmad, line 102: This comment isn't necessary, the code block is simple enough to be understood by itself.

ADmad (commented):
Test case needed for the change.

src/Controller/Component/SearchComponent.php
```diff
@@ -99,8 +99,13 @@ public function beforeRender()
...
+            $model_class = $this->getConfig('modelClass');
+            if (!$model_class) {
```
dereuromark, line 104: Doesnt this mean it skips for the default model, that loadModel() would actually load with empty param?
Wouldnt it be better to check for strict empty string and skip in that case?

```suggestion
            if ($modelClass === '') {
```

dereuromark: Also, why would the component be added to controllers that do not need them?

smladeoye: Loading this component and others in a Parent Controller class that is extended by all other controllers.

smladeoye:
Thanks, will cancel this pr  - simply setting the an empty string modelClass on the controller solves this - the UnexpectedValueException is thrown when trying to load the model and caught in the component.

PR #317, opened 2022-02-16
Remove "Filter" from baked collections.

### LINKED ISSUE #315: Bug/inconvenience when baking FilterCollections
I you bake a FilterCollection through the CLI, e.g.: 

`bin/cake bake filter_collection Users`

this file will be created: 

`class UsersFilterCollection extends FilterCollection`

However, this does not seem to work. It must be `UsersCollection` instead. 

I had quite a hard time finding out why it did not work, until I read the Docs very precisely, where there is a `PostsCollection` and not a `PostsFilterCollection`: 
<https://github.com/FriendsOfCake/search/tree/master/docs#filter-collection-classes>

It would be great, if the Behavior would accept both versions, or if the bake command would create a Filter with proper naming.

ADmad:
Closes #315

### DIFF
--- a/src/Command/Bake/FilterCollectionCommand.php
+++ b/src/Command/Bake/FilterCollectionCommand.php
@@ -167,7 +167,7 @@ class FilterCollectionCommand extends BakeCommand
      */
     public function fileName(string $name): string
     {
-        return $name . 'FilterCollection.php';
+        return $name . 'Collection.php';
     }
 
     /**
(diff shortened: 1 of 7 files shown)


dereuromark (approved):
OK
I would still like it if we used Filter suffix everywhere to make it more clear and searchable what those classes are for.

But since that would be a BC break now, I guess we have to go with this fix at this point.

ADmad:
Ah, you are setting the collection class as default one. Without that it would also require one to specify the `collection` as `foo_filter` instead of just `foo` in the find call.

So it's best to remove "Filter" avoid confusion for users.

PR #318, opened 2022-03-09
Use comparison expression

edoardocavazza:
This PR proposes a refactor of the conditions generated by filters and enables passing fields as expressions.

## Motivations

Right now, filters use strings concatenation to build conditions. This assumes that all input values are string. The `ComparisonExpression` constructor is more versatile, since it also accepts expressions and reduces the parsing overhead of the ORM.

Furthermore, using expressions for fields enables more controlled customisation. For example, we need to perform a case insensitive search in a JSON field. Using expression, we can use `COLLATE` (we have a strongly dependency on mysql) to cast the field to the required encoding:

```php
$this->searchManager()->add('q', 'Search.Like', [
    'before' => true,
    'after' => true,
    'fieldMode' => 'OR',
    'comparison' => 'LIKE',
    'wildcardAny' => '*',
    'wildcardOne' => '?',
    'fields' => [
        new ComparisonExpression('body', 'utf8mb4_general_ci', 'string', 'COLLATE'),
    ],
    'colType' => ['body' => 'string'],
]);
```

### DIFF
--- a/src/Model/Filter/Base.php
+++ b/src/Model/Filter/Base.php
@@ -81,9 +81,7 @@ abstract class Base
 
         if (
             empty($config['fields']) ||
-            (!array_filter($config['fields'], function ($value) {
-                return strlen($value) > 0;
-            }))
+            (!array_filter($config['fields']))
         ) {
             throw new \InvalidArgumentException(
                 'The `field` option is invalid. Expected a non-empty string or array.'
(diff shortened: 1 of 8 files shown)


dereuromark:
We should probably add some examples here in the docs.

ADmad:
The plugin can currently be used with any datasource which implements `\Cake\Datasource\QueryInterface`, for e.g. the `muffin/webservice` plugin. Changing all filters to use `ComparisonExpression` which is part of the `Database` package which break usage with all non-db datasources.

dereuromark:
Could this maybe be its own filter? And we try to support this besides the other cases?

ADmad:
For edge cases like the one shown in the example above one can also use the `Callback` filter and set query conditions as required.

PR #320, opened 2022-03-09
Improve typing

ADmad:


dereuromark:
Looks like prefer lowest has an issue: https://github.com/FriendsOfCake/search/runs/5480841845?check_suite_focus=true#step:7:23
I guess it is not yet working for those cake versions we support?

Issue #136, opened 2016-11-12
filter by belongsToMany


maxmasetti:
hi,
I found this code for filtering by belongsToMany relationship. In the example code, Companies belongsToMany Sectors, in CompaniesTable you can filter by sector with:
```php
    $this->searchManager()
      ->add('sector_id', 'Search.Callback', [
        'callback' => function (Query $query, array $args, Callback $type) {
            return $query
                ->matching('Sectors', function (Query $query) use ($args) {
                    return $query
                        ->where([
                            $this->Sectors->target()->aliasField('id') => $args['sector_id']
                        ]);
                });
        }
      ]);
```
and I would like to share if not known yet.
It could be nice if a sort of shortcut could be used instead of this code, something like
```php
    $this->searchManager()
      ->add('sector_id', 'Search.BelongsToMany', [
        'matching' => 'Sectors',
        'field' => 'sector_id',
      ]);
```
thank you
max

jippi:
PR welcome :)

maxmasetti:
mm, what's "PR" (found 190 defs for it)

tgoeminne:
Hi,

have used this piece of code and it works but only when you use it once time?

How to use it when the belongstomany relation stores extra value?

products_attributes: id product_id attribute_id value

Then want to make multiple filters for this table like:

where attribute_id = 1 & value = 5
where attribute_id = 2 & value = 3

When I copy the add part so there are two callbacks, it does not work.

maxmasetti:
hi,
@tgoeminne: you probably have to customize the inner query, in particular the where clause, but I think there is only one filter field available at that level, so you can build the clause from a tokenized free text passed down. I don't know well how Cake behaves with multiple "matching" on a single many to many associated table, so the problem could be this, in your case, rather than the search plug in: solved the matching separately you can port it back in the filter, as I did for the code I add before.
Hope this helps.

tgoeminne:
->add('Color', 'Search.Callback', [
        		'callback' => function ($query, $args, $filter) {
       			return $query
        		->matching('Attributes', function ($query) use ($args) {
        			return $query
        			->where([['ProductsAttributes.eav IN' => $args['Color'],'ProductsAttributes.attribute_id' => '20']]);
        		});
        	}
        ])
        ->add('Brand', 'Search.Callback', [
            'callback' => function ($query, $args, $filter) {
            return $query
            ->matching('Attributes', function ($query) use ($args) {
                return $query
[...]

PR #323, opened 2022-05-28
add Matching filter

LordSimal:
Fixes https://github.com/FriendsOfCake/search/issues/136

Currrently this HABTM filter only allows for `OR` connected IDs, so its basically a
```
WHERE MyTable.id IN [1,2,3,4]
```

If we wanted to have a `AND` connected filter possibility as well I could get that to work if you want.
My apporach would be sub-selects for each ID that is given. Or is there a better way to handle that?

<s>Actually, renaming that Filter to `Matching` would be more appropriate I think...</s> done!

### DIFF
--- a/src/Model/Filter/HABTM.php
+++ b/src/Model/Filter/HABTM.php
@@ -8,7 +8,7 @@ use Cake\ORM\Query;
 class HABTM extends Base
 {
     /**
-     * Modify query using callback.
+     * Modify query to filter a HABTM association.
      *
      * @return bool
      */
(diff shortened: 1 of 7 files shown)


ndm2:
I'm not sure if such a filter is a good idea, as it's hard to make it work properly with all the caveats surrounding filtering associations. There would at least need to be grouping, as otherwise this can easily cause duplicates to be returned, and once you use grouping, you're opening a big can of worms with strict grouping, and while using `innerJoinWith()` (which I'd always recommend over `matching()`) can avoid some of the problems caused by the inclusion of all the associated table's fields in the select list, it can still easily break.

Doing an `AND` comparison can be quite tricky, there's all sorts of solutions, but you'll most likely never find one that will satisfy everyone, things will break, or performance will be bad, or whatever. Personally I often use `HAVING` to count the unique associated records, eg if you have 2 condition values to match against, you require the groups to have two unique values of that column, like

```sql
WHERE Tags.id IN (1, 2)
GROUP BY Articles.id
HAVING COUNT(DISTINCT Tags.id) = 2
```

LordSimal:
So in the end everyone still needs to use the callback filter to create their own filter based upon their specific HABTM association because there is no straight forward solution... I had a feeling it will come to this way. Thanks anyway for the info @ndm2 🙇🏻

ndm2:
I'd say the best for now would be to update the docs with some examples of how to filter associations using the callback filter.

PR #324, opened 2022-05-28
docs: split up into multiple files and clarify examples

LordSimal:
Relates to https://github.com/FriendsOfCake/search/pull/323 and https://github.com/FriendsOfCake/search/issues/136


raul338:
This sounds like I used on #249 
It can be done using the `value` filter

```php
$manager->value('category_id', [
    'multiValue' => true,
    'multiValueSeparator' => ',',
    'field' => ['Categories.id'],
    'beforeProcess' => function (Query $query, array $args, $filter) {
        $query->innerJoinWith('Categories');
    },
]);
```

so in the URL you can pass `category_id=1,2,3,4` (and I belive an array too)

LordSimal:
Indeed this does work as well.
But in my opinion your solution is thinking quite "out of the box" just with the docs given currently.
Especially for newer people who are not that accustomed to the framework.
Or is there a benefit to your approach?

raul338:
My code is combining two filter config in the docs [beforeProcess](https://github.com/FriendsOfCake/search/blob/1994e979a5c669f3ea9111026e82ff1636bfff1e/docs/README.md?plain=1#L350) and [Value Filter](https://github.com/FriendsOfCake/search/blob/master/docs/README.md#value)

The code serves as an example of the value filter.
For the callback I would suggest the example would make use of the other parameters

LordSimal:
I didn't mean that I don't understand your example 😉 
I was just saying that especially for newcomers a little code block explaining how to filter HABTM associations works can go along way instead of having them reading through all filters and options to let them figure it out themselves.

Since the docs/README.md is already quite big I wouldn't mind splitting it up into several pages, so something like
- Basic Configuration
- All default filters incl. examples
- Additional docs (like everything after the Optional Fields in the current Readme)

raul338:
I agree it would be better to show examples, or repeat titles/sections like it's done on the crud plugin documentation

ndm2:
> Since the docs/README.md is already quite big I wouldn't mind splitting it up into several pages

All the options on themselves can be quite overwhelming, so a restructuring of the docs would be quite nice I think, dedicated sections for all filters, each with an example would be very helpful especially for new users.

LordSimal:
This is my first version of the split up doc including the more simpler, explicit filter examples:
https://github.com/LordSimal/search/tree/how-to-use-closure/docs

LordSimal:
So are there any other objections if my approach of splitting up the docs is good/bad?

docs/filters-and-examples.md
```diff
@@ -0,0 +1,349 @@
+# Filters
+
+The Search plugin comes with a set of predefined search filters that allow you to
+easily create the search results you need. Use:
+
+----------
+
+`Value` to limit results to exact matches
+<details>
```
ADmad, line 9: The details are collapsed by default and I don't like the fact that an additional click is required to view the code e.g.

LordSimal: since most of these code snippets are only one liners I can actually remove that collapsible element.

Issue #329, opened 2022-10-14
BC issue


dereuromark:
https://github.com/FriendsOfCake/search/pull/320 seems to be not fully BC
Controllers are now broken in projects that update to latest plugin version

> You must provide an $alias or set the $defaultTable property.

loadModel() didnt need an argument
fetchTable() now does, ot should default to the default model (`$defaultTable`) of the controller here.
It just is not available via public API it seems, so the component cannot access it.

Could we use reflection to shim this?

ADmad:
`$defaultTable` does get set to the same values as `$modelClass` https://github.com/cakephp/cakephp/blob/bae80000d28106aaf7bca509db02b556340f3ea2/src/Controller/Controller.php#L222

Have you explicitly set the `$modelClass` property in the controller?

dereuromark:
No, since this was not needed
having to do so now would need to touch thousands of controllers. thats not ideal.

ADmad:
Not saying you need to do that. Just trying to figure what could be causing the problem. As shown in the code linked above `$defaultTable` should already have a value set through the controller's constructor, so I don't see how/why you get that error.

dereuromark:
Interesting, could that project have maybe a missing parent call on any specific controller callback? I will check.
The linked PR fixed it though for us.

dereuromark:
OK, all AppController and specific controller callbacks call parent in the first line.
That cannot be it.

ADmad:
Let's 1st figure out why the error is occurring before patching the code.

The property gets set in the constructor itself so what callbacks do isn't relevant. Have you overridden the controller's constructor?

dereuromark:
No, only initialize()
```php
    public function initialize(): void
    {
        parent::initialize();

        $this->loadComponent('RequestHandler');
        ...
}
```

ADmad:
What does `debug($this->defaultTable)` in `initialize()` give you?

dereuromark:
It is null, but modelClass is set

Also here it is null in Core Controller
```
        if ($this->defaultTable !== null) {
            $this->modelClass = $this->defaultTable;
        }
```

Where is $defaultTable supposed to be "set"?

dereuromark:
I think I understand the issue
The project does not specify any defaultTable (neither should it need to) in any controller, so only modelClass is set in Controller in core
As such it can never have it populated later since it is never written to.

ADmad:
> It is null, but modelClass is set

[...]

PR #330, opened 2022-10-14
Fix BC issue for legacy projects

dereuromark:
Fixes https://github.com/FriendsOfCake/search/issues/329

Avoids having to fix all controllers here.


src/Controller/Component/SearchComponent.php
```diff
@@ -103,8 +104,16 @@ public function beforeRender()
             if (method_exists($controller, 'fetchTable')) {
                 $method = 'fetchTable';
             }
+            $modelClassName = $this->getConfig('modelClass');
+            if (!$modelClassName && $method === 'fetchTable') {
```
dereuromark, line 108: We only invoke this in this very specific instance to keep BC, nothing else should be affected, like happy normal path.

PR #332, opened 2023-06-08
cake5: update to PHPUnit 10 and use named arguments for findSearch

LordSimal:


src/Model/SearchTrait.php
```diff
@@ -45,22 +44,24 @@ trait SearchTrait
...
+     * @param string $collection Filter collection name.
      * @return \Cake\Datasource\QueryInterface The Query object used in pagination.
      * @throws \Exception When missing search arguments.
      */
-    public function findSearch(QueryInterface $query, array $options): QueryInterface
-    {
-        if (!isset($options['search'])) {
+    public function findSearch(
+        QueryInterface $query,
+        array $search = [],
```
ADmad, line 54: Remove the default value. Calling this method without search params would be pointless.

src/Model/SearchTrait.php
```diff
@@ -45,22 +44,24 @@ trait SearchTrait
...
+        string $collection = Manager::DEFAULT_COLLECTION
+    ): QueryInterface {
+        if (!$search) {
```
ADmad, line 57: This is no longer necessary once default value is removed.

tests/TestCase/Model/Behavior/SearchBehaviorTest.php
```diff
@@ -318,7 +319,7 @@ public function collectionFinderProvider()
     public function testFindSearchException()
```
ADmad, line 319: This test case can be dropped.

tests/TestCase/Model/SearchTraitTest.php
```diff
@@ -83,7 +83,7 @@ public function testFinder()
     public function testFindSearchException()
```
ADmad, line 83: This test can be dropped.

ADmad (commented):
> should I do the stan fix here as well?

Yes please.

tests/TestCase/Model/SearchTraitTest.php
LordSimal: I wasn't sure if we wanna keep the manually thrown exception or go with the argument error.
But I aggree that providing a search named param is required 👍🏻

ADmad: I went through all this trouble of having finder with typed args so that we can just rely on the language and not have to add checks myself :laughing:.

PR #340, opened 2024-01-04
More clear feedback on typo in model class setup.

dereuromark:
I had a typo in the model class `BitmaskedRecords` and things appeared to work fine, even though the model was never loaded.

```
		$this->loadComponent('Search.Search', [
			'actions' => ['bitmaskSearch'],
			'queryStringWhitelist' => ['type'],
			'modelClass' => 'Sandbox.BitmaskedRecords',
		]);
```
and I didnt notice that things are wrong, only when looking closer to the reason why the reset button would not show.
With this info it should be clear that a configured value would need to exist or should otherwise be left null/empty.

### DIFF
--- a/src/Controller/Component/SearchComponent.php
+++ b/src/Controller/Component/SearchComponent.php
@@ -108,6 +108,12 @@ class SearchComponent extends Component
         }
 
         if (!$model->behaviors()->has('Search')) {
+            if ($this->getConfig('modelClass')) {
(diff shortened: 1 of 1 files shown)


ADmad:
If the issue was about the model not being found/loaded then changes should be done to the try/catch block above which has the `fetchTable()` call, not when checking for the behavior.

Issue #341, opened 2024-02-24
Casting needed for Finder filter


dereuromark:
Currently it has the options

        'map' => [],
        'options' => [],
        
But I get

> findUser(): Argument #2 ($uid) must be of type int, string given

With URL

> /admin/profiles?user_id=115

I guess the string '115' is never casted to int, the custom finder however is

    public function findUser(SelectQuery $query, int $uid)

Sure, we can workaround it using

    public function findUser(SelectQuery $query, int|string $uid)
    
But it seems a bit less clean than allowing

    'cast' => [
        'field_name' => 'int',
    ],
    
or alike as additional option to cast the value accordingly before passing it on?

Issue #338, opened 2023-12-04
5.x: Search exception


dereuromark:
I have params of 

    'city_id' => (int) 76,

Which then triggers exception

> Search\Model\Filter\Base::passedValue(): Return value must be of type array|string|null, int returned
```php
    /**
     * @return array|string|null
     */
    protected function passedValue(): string|array|null
```

I wonder if this should be also allowing other scalar values actually?
As it seems hard to control the exact types going into the search here IMO

dereuromark:
I worked around it for now by casting

    $search['city_id'] = (string)$city->id;

before passing those further on.

ADmad:
How/why is `city_id` an `int`?  Passed values normally come request query string which are always strings.

dereuromark:
The data doesnt completely come from the request, thats why
It seems strange to have to take a primary key int, make it string only to become int afterwards again, though.

ADmad:
If we are going to widen the type then let's just make it `mixed` instead of limit to scalars and array. This value is eventually going to be used in a query which does accept various types of objects depending on the field type. Also `Base::value()` which calls `passedValue()` already has `mixed` return type.

dereuromark:
Turns out, the wrapping method already expects mixed.

    /**
     * Get the value of the "name" from HTTP GET arguments.
     *
     * @return mixed
     */
    public function value(): mixed
    {
        $value = $this->_config['defaultValue'];

        $passedValue = $this->passedValue();
        if ($passedValue === null) {
            return $value;
        }

        return $passedValue;
    }
    
So, yeah, mixed is the correct way to go

PR #342, opened 2024-02-24
Add casting for Finder.

dereuromark:
With stricter method signatures this becomes necessary for Finder usage.

Resolves 
- https://github.com/FriendsOfCake/search/issues/338
- https://github.com/FriendsOfCake/search/issues/341

We can release a new minor 7.1.0

### DIFF
--- a/src/Model/Filter/Finder.php
+++ b/src/Model/Filter/Finder.php
@@ -6,7 +6,7 @@ namespace Search\Model\Filter;
 class Finder extends Base
 {
     /**
-     * @var array
+     * @var array<string, mixed>
      */
     protected array $_defaultConfig = [
         'map' => [],
--- a/src/Model/Filter/Finder.php
+++ b/src/Model/Filter/Finder.php
@@ -3,6 +3,8 @@ declare(strict_types=1);
 
 namespace Search\Model\Filter;
 
+use Closure;
+
 class Finder extends Base
 {
     /**
@@ -45,7 +47,7 @@ class Finder extends Base
                 continue;
             }
 
-            if (is_callable($toType)) {
+            if ($toType instanceof Closure) {
                 $value = $toType($value);
             } else {
                 settype($value, $toType);
--- a/src/Model/Filter/Finder.php
+++ b/src/Model/Filter/Finder.php
@@ -11,6 +11,7 @@ class Finder extends Base
     protected array $_defaultConfig = [
         'map' => [],
         'options' => [],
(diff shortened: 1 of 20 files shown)


src/Model/Filter/Finder.php
```diff
@@ -37,6 +38,18 @@ public function process(): bool
         foreach ($map as $to => $from) {
             $args[$to] = $args[$from] ?? null;
         }
+        $casts = $this->getConfig('cast');
+        foreach ($casts as $field => $toType) {
+            $value = $args[$field] ?? null;
+
+            if (is_callable($toType)) {
```
ADmad, line 45: The type setting and calling the callable should be skipped if the value is null, otherwise the null will get cast to integer 0 for e.g. There should also be a test for it.

dereuromark: I just checked and added tests for both normal and callable using null input.
And it works as expected. How would we get an issue that you describe?

ADmad: > I also added tests for when empty (request) strings need to be mapped to nullish value for ?type args.

I don't see that being useful in practice as `Processor::_extractParams()` filters out all falsey values by default.

The change I was suggesting was this:

```php
if ($value !== null) {
    if (is_callable($toType)) {
        $value = $toType($value);
    } else {
        settype($value, $toType);
    }
}

Skipping the casting all together if no value has been provided aka `null`.
```

ADmad: The various tests cases you have setting which have `$filter->setArgs(['foo' => null]);` have negligible practical value and filtering records with a a column matching null / empty string is rarely needed. For those rare cases people can just use the `Callback` filter and even for that to work they will have to first change the empty values filtering.

dereuromark:
Adjusted as per feedback

src/Model/Filter/Finder.php
```diff
@@ -37,6 +38,21 @@ public function process(): bool
...
             $args[$to] = $args[$from] ?? null;
         }
+        $casts = $this->getConfig('cast');
+        foreach ($casts as $field => $toType) {
+            $value = $args[$field] ?? null;
+            if ($value === null) {
+                continue;
+            }
+
+            if (is_callable($toType)) {
```
ADmad, line 48:
```suggestion
            if ($toType instanceof \Closure) {
```
I highly dislike `is_callable()` and doubt anyone will use anything besides a closure here.

PR #345, opened 2025-04-04
multiValueExactMatching feature

dereuromark:
This enables exact quote matches
```php
        $filter->setConfig('multiValueSeparator', ' ');
        $filter->setConfig('multiValueExactMatching', true);

        $filter->setArgs(['fields' => 'value1 "value2 and 3" value4']);
        $this->assertEquals(['value1', 'value2 and 3', 'value4'], $filter->value());

        $filter->setConfig('multiValueSeparator', ' ');
        $filter->setConfig('multiValueExactMatching', '*');

        $filter->setArgs(['fields' => 'value1 *value2 and 3* value4']);
        $this->assertEquals(['value1', 'value2 and 3', 'value4'], $filter->value());
```

Changes are in the 1st commit.
The 2nd is just to make CS happy again.

### DIFF
--- a/src/Model/Filter/Base.php
+++ b/src/Model/Filter/Base.php
@@ -238,9 +238,10 @@ abstract class Base
 
         $terms = [];
 
+        $escapedQuoteChar = preg_quote($quoteChar, '/');
         // Match quoted phrases and unquoted words
         preg_match_all(
-            '/' . preg_quote($quoteChar) . '([^' . preg_quote($quoteChar) . ']+)' . preg_quote($quoteChar) . '|\S+/',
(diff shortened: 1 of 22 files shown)


src/Model/Filter/Base.php
```diff
@@ -208,12 +209,55 @@ protected function passedValue(): mixed
...
         }
 
         return $value;
     }
 
+    /**
+     * @param string $input
+     * @return array<string>
+     */
+    protected function parseSearchTerms(string $input): array
```
ADmad, line 226: Why is all this parsing code in `Base` when `multiValueSeparator` and `multiValueExactMatching` options are only relevant for `Value` filter?

ADmad: I guess I can look into any potential refactoring after merging this.

Issue #347, opened 2025-04-17
Issue: Filter range of two combinated fields


dereuromark:
https://github.com/dereuromark/cakephp-sandbox/pull/91

This works, but the isSearch is false and as such the reset button doesnt work..
Any idea why it would think that?
```php
		$searchManager
			->like('title', ['before' => true, 'after' => true])
			->callback('price_max', [
				'callback' => function (SelectQuery $query, array $args, $filter) {
					$min = (int)$args['price_min'];
					$max = (int)($args['price_max'] ?? null);

					if (!$min && !$max || $max < $min) {
						return false;
					}

					return true;
				},
			])
			->callback('price_min', [
				'callback' => function (SelectQuery $query, array $args, $filter) {
					$min = (int)$args['price_min'];
					$max = (int)($args['price_max'] ?? null);
					if (!$min && !$max || $max < $min) {
						return false;
					}

					$query->where(['price >=' => $min, 'price <=' => $max]);

					return true;
				},
			]);
```

Demo: https://sandbox.dereuromark.de/sandbox/search-examples/range

dereuromark:
My bad, I found the issue

but in order for isSearch to not always be a active, I need to remove them dynamically it seems:
```php
		// Trick to avoid it being counted as isSearch for default range
		$queryParams = $this->request->getQuery();
		if ($min && (int)$this->request->getQuery('price_min') === (int)$min->price && $max && (int)$this->request->getQuery('price_max') === (int)ceil((float)$max->price)) {
			unset($queryParams['price_min']);
			unset($queryParams['price_max']);
		}

		$query = $productsTable->find('search', search: $queryParams);
```

Or is there a better way?

dereuromark:
Note that in my case I am dynamically adding min/max to the JS to define the slider range:

Adjust code style settings
```php
/** @var \Sandbox\Model\Entity\Product|null $min */
$min = $this->fetchTable('Sandbox.Products')->find()->orderByAsc('price')->first();
/** @var \Sandbox\Model\Entity\Product|null $max */
$max = $this->fetchTable('Sandbox.Products')->find()->orderByDesc('price')->first();
```

and
```
slider.slider({
	range: true,
	values: [val0, val1],
	min: <?php echo (int)$min->price; ?>,
	step: <?php echo (int)(ceil($max->price) / 100); ?>,
	minRange: <?php echo (int)(ceil($max->price) / 10); ?>,
[...]

PR #348, opened 2025-04-17
Fix up empty via callable for more dynamic use cases.

dereuromark:
Resolves https://github.com/FriendsOfCake/search/issues/347
```php
	$callable = function ($value, array $params) use ($minPrice, $maxPrice): bool {
		$minValue = $minPrice ? (int)$minPrice->price : 0;
		$maxValue = $maxPrice ? (int)ceil((float)$maxPrice->price) : 0;

		if (!$minPrice && !$maxPrice) {
			return true;
		}
		if (empty($params['price_min']) || empty($params['price_max'])) {
			return true;
		}

		if ((string)$minValue === $params['price_min'] && (string)$maxValue === $params['price_max']) {
			return true;
		}

		return false;
	};
	$this->Search->setConfig('emptyValues', [
		'price_min' => $callable,
		'price_max' => $callable,
	]);
```	

With this I dont need hacks anymore in the controller actions (that still dont properly work fully)

### DIFF
--- a/src/Controller/Component/SearchComponent.php
+++ b/src/Controller/Component/SearchComponent.php
@@ -28,7 +28,7 @@ class SearchComponent extends Component
      *   Defaults to the Paginator `'sort'`, `'direction'` and `'limit'` ones.
      * - `queryStringBlacklist` : An array of form fields that should not end up in the query.
      * - `emptyValues` : A map of fields and their values to be considered empty
-     *   (will not be passed along in the URL).
+     *   (will not be passed along in the URL). Use callables for more control (return true for "empty").
      * - `modelClass` : Configure the controller's modelClass to be used for the query, used to
      *   populate the _isSearch view variable to allow for a reset button, for example.
      *   Set to false to disable the auto-setting of the view variable.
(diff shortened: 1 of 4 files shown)


src/Controller/Component/SearchComponent.php
```diff
@@ -147,6 +147,15 @@ protected function _filterParams(): array
                 continue;
             }
 
+            if (is_callable($value)) {
```
ADmad, line 150: Lets limit this to a `\Closure` instance.

dereuromark: Is that better?
I just thought since we already use
- `Callback` to produce results using your own custom callable function

it seems to be consistent for now.

ADmad: Using `Closure` whenever possible is better since callable is a very wide type in PHP. For empty values checking it's highly unlikely anyone will need something more than a short anonymous function.

ADmad: Also the availability of [first class callable](https://www.php.net/manual/en/functions.first_class_callable_syntax.php) makes it trivial to turn any non callable into a Closure instance.

dereuromark: Done :)

PR #350, opened 2025-05-01
Fixed deprecation warning in CakePHP 5.2

bakewizard:
This PR fixes a deprecation warning that appears when using the **FriendsOfCake/search** plugin with CakePHP 5.2.

### DIFF
--- a/src/Controller/Component/SearchComponent.php
+++ b/src/Controller/Component/SearchComponent.php
@@ -5,7 +5,6 @@ namespace Search\Controller\Component;
 
 use Cake\Controller\Component;
 use Cake\Event\EventInterface;
-use Cake\Http\Response;
 use Cake\Utility\Hash;
 use Closure;
 use UnexpectedValueException;
@@ -66,13 +65,13 @@ class SearchComponent extends Component
      * Checks if the current request has posted data and redirects the users
      * to the same action after converting the post data into GET params
      *
-     * @param \Cake\Event\EventInterface $event The event instance.
-     * @return \Cake\Http\Response|null
+     * @param \Cake\Event\EventInterface $event Event instance
+     * @return void
      */
-    public function startup(EventInterface $event): ?Response
+    public function startup(EventInterface $event): void
     {
         if (!$this->getController()->getRequest()->is('post') || !$this->_isSearchAction()) {
-            return null;
+            return;
         }
 
         $url = $this->getController()->getRequest()->getPath();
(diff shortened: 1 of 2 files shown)


src/Controller/Component/SearchComponent.php
```diff
@@ -65,9 +66,10 @@ public function implementedEvents(): array
      * Checks if the current request has posted data and redirects the users
      * to the same action after converting the post data into GET params
      *
+     * @param \Cake\Event\EventInterface $event The event instance.
      * @return \Cake\Http\Response|null
      */
-    public function startup(): ?Response
+    public function startup(EventInterface $event): ?Response
```
dereuromark, line 72: If we add a param here, I wonder if we might as well take the return type away and make it void?
Or is that too BC breaking?

Idieally, no one should be using this component via extension..

src/Controller/Component/SearchComponent.php
```diff
@@ -171,4 +173,4 @@ protected function _filterParams(): array
 
         return $params;
     }
-}
+}
```
dereuromark, line 176: you need to set up your IDE correctly, it should keep trailing newlines on files.

src/Controller/Component/SearchComponent.php
dereuromark: If this is already working maybe we just keep things as is now.
Next major can clean up return type signature :)

dereuromark: But if you keep the return type here, you need to "return null" at the end of the method IMO.

bakewizard: You're right. Its return type should be void

src/Controller/Component/SearchComponent.php
bakewizard: Got it, I'll fix my IDE settings.

src/Controller/Component/SearchComponent.php
dereuromark: I think we can fix all up and release a new minor with changelogs.

dereuromark: Given the failed builds in CI I guess the tests might also need adjustment.

src/Controller/Component/SearchComponent.php
dereuromark: any update on the fixed PR?

PR #371, opened 2026-01-28
Add strictMode config for misconfigured SearchComponent

dereuromark:
## Summary

- Adds a `strictMode` config option to `SearchComponent` (default `false`) that controls behavior when the model cannot be fetched or the Search behavior is not loaded
- With `strictMode` disabled (default): silently skips, preserving the existing behavior that allows loading the component in `AppController` where not all controllers have a searchable table
- With `strictMode` enabled: throws `RuntimeException` with a clear message identifying the controller, action, and what went wrong — catching misconfiguration early

## Changes

- `src/Controller/Component/SearchComponent.php`:
  - Added `strictMode` config option (default `false`)
  - When `strictMode` is true and `fetchTable()` fails: throws `RuntimeException` with the original exception chained
  - When `strictMode` is true and Search behavior is missing: throws `RuntimeException`
  - When `strictMode` is false: silently returns (no change from current master behavior)
- `tests/TestCase/Controller/Component/SearchComponentTest.php`:
  - Added `testBeforeRenderWithoutModelSilentlySkips` — verifies default silent behavior
  - Added `testBeforeRenderWithoutBehaviorSilentlySkips` — verifies default silent behavior
  - Added `testStrictModeThrowsOnMissingModel` — verifies exception with strictMode
  - Added `testStrictModeThrowsOnMissingBehavior` — verifies exception with strictMode

## Migration path

In a future major version, `strictMode` could become `true` by default, with users opting out via `'strictMode' => false` for the AppController use case.

### DIFF
--- a/src/Controller/Component/SearchComponent.php
+++ b/src/Controller/Component/SearchComponent.php
@@ -167,10 +167,32 @@ class SearchComponent extends Component
              */
             $model = $controller->fetchTable($this->getConfig('modelClass'));
         } catch (UnexpectedValueException $e) {
+            deprecationWarning(
+                '5.0.0',
+                sprintf(
+                    'SearchComponent on `%s::%s()` could not load table: %s. '
+                    . 'Set the `modelClass` config option to the correct table class.',
+                    get_class($controller),
+                    $controller->getRequest()->getParam('action'),
+                    $e->getMessage(),
+                ),
+            );
+
             return;
         }
 
         if (!$model->behaviors()->has('Search')) {
+            deprecationWarning(
+                '5.0.0',
+                sprintf(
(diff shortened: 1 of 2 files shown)


src/Controller/Component/SearchComponent.php
```diff
@@ -167,10 +168,32 @@ public function beforeRender(): void
              */
             $model = $controller->fetchTable($this->getConfig('modelClass'));
         } catch (UnexpectedValueException $e) {
+            deprecationWarning(
+                '5.0.0',
```
ADmad: Shouldn't this be the next minor of the plugin? The issue is unrelated to CakePHP core.

ADmad (approved):

PR #373, opened 2026-02-09
Add Mapped filter for default values without triggering isSearch

dereuromark:
## Summary

This adds a new `mapped()` filter type that solves a common pattern: applying a filter by default without triggering `isSearch()` (which controls the Reset button visibility).

### Use Case

You want:
- A filter to apply by default (e.g., show only "enabled" records)
- The Reset button to NOT appear when using the default
- An "All" option that removes the filter and DOES show Reset
- Explicit non-default options that also show Reset

### API

```php
// Boolean example
$this->mapped('enabled', [
    'map' => ['' => true, '0' => false, '-1' => null],
    'default' => '',
]);

// Enum example (unmapped values pass through)
$this->mapped('status', [
    'map' => ['' => 'pending', '-1' => null],
    'default' => '',
]);
// 'active', 'completed' etc. pass through as-is
```

### Features

- **Map keys** are form values, **map values** are the condition to apply
- `null` in map means "no filter condition" (show all)
- `default` key specifies which value doesn't trigger `isSearch()`
- Non-empty values not in map pass through directly as filter condition
- Automatically sets `alwaysRun=true` and `filterEmpty=false`

### Form Template

```php
<?= $this->Form->control('enabled', [
    'options' => [
        '' => __('Enabled'),      // Default
        '0' => __('Disabled'),
        '-1' => __('All'),
    ],
]) ?>

<?php if ($this->Search->isSearch()) { ?>
    <?= $this->Search->resetLink(__('Reset')) ?>
<?php } ?>
```

## Changes

- New `Mapped` filter class
- Added `mapped()` method to `FilterMethodsTrait`
- Unit tests for all scenarios

As a result my App code would significantly clean up:
```diff
-			->callback('enabled', [
-				'callback' => function (SelectQuery $query, array $args, $filter): bool {
-					$value = $args['enabled'] ?? null;
-
-					if ($value === '-1') {
-						// All - user explicitly chose "no filter"
-						return true;
-					}
-					if ($value === '0') {
-						// Disabled - user explicitly chose this
-						$query->where(['DiscoveryProfiles.enabled' => false]);
-
-						return true;
-					}
-
-					// Default (empty or null) = enabled
-					$query->where(['DiscoveryProfiles.enabled' => true]);
-
-					return false; // Default state, don't count as active search
-				},
-				'alwaysRun' => true,
-				'filterEmpty' => false,
+			->mapped('enabled', [
+				'fields' => 'DiscoveryProfiles.enabled',
+				'map' => [
+					'' => true,      // Default: show enabled
+					'-1' => null,    // Show all (no filter)
+				],
+				'default' => '',     // Empty string is the default (doesn't trigger isSearch)
 			])
```
0 (false) is passed through, no need to define.

### DIFF
--- a/docs/filters-and-examples.md
+++ b/docs/filters-and-examples.md
@@ -78,6 +78,34 @@ $searchManager->callback('category_id', [
 
 ----------
 
+`Mapped` to map form values to filter conditions with support for defaults that don't
+trigger `isSearch()`. Useful for filters with a default value (e.g., "show enabled by
+default") where you don't want the Reset button to appear.
+
+```php
+// Boolean example: default to enabled=true, allow showing all with -1
+$searchManager->mapped('enabled', [
+    'map' => ['' => true, '0' => false, '-1' => null],
+    'default' => '',
+]);
+
+// Enum example: default to 'pending' status, unmapped values pass through
+$searchManager->mapped('status', [
+    'map' => ['' => 'pending', '-1' => null],
+    'default' => '',
+]);
+// Values like 'active', 'completed' pass through directly as filter condition
+```
+
+Key features:
+- Map keys are form values, map values are the condition to apply
+- `null` in map means "no filter condition" (show all)
+- `default` key specifies which value doesn't trigger `isSearch()`
+- Non-empty values not in map pass through directly as filter condition
+- Sets `alwaysRun: true` and `filterEmpty: false` by default
+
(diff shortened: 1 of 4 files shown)


dereuromark:
@ADmad Can this move forward?

ADmad (approved):

PR #374, opened 2026-05-11
Fix extraParams leak, Callback null contract, Like driver detection

### LINKED ISSUE #341: Casting needed for Finder filter

### LINKED ISSUE #351: Allow more easy "empty value" search
Based on https://discourse.cakephp.org/t/paginator-for-cakephp5-x-the-url-on-the-next-page-has-removed-fields-with-empty-values/12568/2

Currently, one can use emptyValues to filter out certain empty values from the URL to be kept.

The opposite, to keep them in to be filtered by is a bit more tricky.
I usually use the trick of -1 and manually change that back to empty string or null in the filter, requiring custom ones then.

It would be nice if there was a way to pass through a not-empty-string "empty" value and auto transform that back to the fitting value inside?
Could we add a "emptyValue" to the Filter?
And if that is set, we convert back to null?

E.g.
/users?email=-&page=2
which then is kept going through pagination (it internally uses array_filter to remove empty values, thus the `-` char)

And the `-` becomes then null inside the filter, to be processed further again.

Any issues with allowing this to be used opt-in?

dereuromark:
## Summary

Three small independent fixes, one commit per fix so they can be reverted independently if needed.

### 1. `Processor::process()` — extraParams leak across filters

`$filterParams` was mutated in place inside the filter loop, so any filter that declared `extraParams => [...]` polluted that map for every filter that ran after it. A filter that did NOT request `extra_field` could still observe it via `getArgs()['extra_field']`, breaking the documented per-filter scoping. Build a per-iteration `$currentParams` view instead.

Regression test: `SearchBehaviorTest::testExtraParamsDoNotLeakBetweenFilters` — declares two callbacks, only the first requests `extra_field`, asserts the second's `args` is clean. Fails before the fix.

### 2. `Callback::process()` — null return contract

The docs (`docs/filters-and-examples.md:67`) and the per-plugin maintainer rule say callbacks must return `bool` to drive `isSearch()`. The implementation did `call_user_func(...) ?? true`, silently coercing a missing return into `isSearch=true`. The README/example at `docs/basic-configuration.md:46` itself omits the return statement, codifying the bug.

Soft-deprecation strategy: keep the historic `null → true` behavior for one minor (existing apps still work), but emit `deprecationWarning('7.9.0', ...)` pointing at the offending callback by name so users can fix it. Next minor can hard-fail.

Existing `CallbackTest::testProcess` updated to `return true` explicitly. New `testProcessNullReturnIsDeprecated` asserts the deprecation fires and the return value is still `true`.

### 3. `Like::_setEscaper()` — brittle driver detection + sticky cache

Two problems in one block:

- **Detection** was `substr_compare(get_class($driver), 'Sqlserver', -9) === 0`. Any subclassed or wrapped Sqlserver driver (custom App driver class) is misdetected as default.
- **Stickiness**: the resolved class was written back into the filter's config, so a reused filter instance kept the first detected escaper even when the next query ran on a different connection.

Switched to `instanceof` against the driver instance, added a `Postgres` branch with a new `PostgresEscaper` class that currently inherits from `DefaultEscaper` (a stub for now — driver-specific wildcard rules can diverge later without breaking the public API), and removed the `setConfig('escaper', ...)` write-back so each `_setEscaper()` call resolves afresh.

Three regression tests, all using anonymous subclasses of the actual Cake driver classes + a stub connection that injects them (avoids needing a real Postgres/SQL Server server in CI):

- `testEscaperSelectionForSqlServerDriverSubclass`,
- `testEscaperSelectionForPostgresDriver`,
- `testEscaperResolvesAfreshPerQuery`.

## Gates

All green locally: `phpunit` (156 tests, 337 assertions, 2 pre-existing skips), `phpstan` level 8 (clean), `phpcs` (clean).

## Notes

- All three fixes target `master` (CakePHP 5). Suggested release target: `7.9.0` — the Callback deprecation is the only behavior-visible change and naturally fits a minor bump.
- Out of scope (next PR): pluggable escaper resolver — a config knob letting apps register `driver => escaperClass` mappings without subclassing the filter. Builds cleanly on top of the per-query resolution from this PR.
- Also out of scope: `Finder::process()` argument filtering (closed #341), `Value` `IS NULL` support (closed #351), migrating tests to fixture-factories.

### DIFF
--- a/src/Model/Filter/Callback.php
+++ b/src/Model/Filter/Callback.php
@@ -40,7 +40,7 @@ class Callback extends Base
                 '7.9.0',
                 sprintf(
                     'Callback filter `%s` returned null; callbacks must return bool to control isSearch(). '
-                    . 'In a future version returning null/void will throw.',
+                    . 'In a future version returning null/void will throw an error.',
                     $this->name(),
                 ),
             );
--- a/src/Model/Filter/Callback.php
+++ b/src/Model/Filter/Callback.php
@@ -3,6 +3,8 @@ declare(strict_types=1);
 
 namespace Search\Model\Filter;
 
+use function Cake\Core\deprecationWarning;
+
 class Callback extends Base
 {
     /**
@@ -17,15 +19,35 @@ class Callback extends Base
     /**
      * Modify query using callback.
      *
+     * The callback must return a `bool` indicating whether it modified the
+     * query (controls `isSearch()`). Returning `null`/void is supported for
+     * backwards compatibility but is deprecated and will throw in a future
+     * version.
+     *
      * @return bool
      */
     public function process(): bool
     {
-        return call_user_func(
+        $result = call_user_func(
             $this->getConfig('callback'),
             $this->getQuery(),
             $this->getArgs(),
             $this,
-        ) ?? true;
+        );
+
+        if ($result === null) {
+            deprecationWarning(
+                '7.9.0',
+                sprintf(
+                    'Callback filter `%s` returned null; callbacks must return bool to control isSearch(). '
+                    . 'In a future version returning null/void will throw.',
+                    $this->name(),
+                ),
+            );
+
+            return true;
+        }
+
+        return $result;
     }
 }
--- a/docs/filters-and-examples.md
+++ b/docs/filters-and-examples.md
@@ -288,12 +288,44 @@ The following options are supported by all filters except `Callback` and `Finder
   (`_`) is being escaped in case used the search term.
 
 - `escaper` (`string`, default to `null`) Defines the escaper that should
-  escape `%` and `_`. If no escaper is set (`escapeDriver => 'null'`) the escaper
-  is set by database driver. If the driver is `Sqlserver` the `SqlserverEscaper`
-  is used (escaping `%` to `[%]` and `_` to `[_]`). In all other cases the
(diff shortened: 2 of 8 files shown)


src/Model/Filter/Callback.php
```diff
@@ -17,15 +19,35 @@ class Callback extends Base
...
             $this,
-        ) ?? true;
+        );
+
+        if ($result === null) {
+            deprecationWarning(
+                '7.9.0',
+                sprintf(
+                    'Callback filter `%s` returned null; callbacks must return bool to control isSearch(). '
+                    . 'In a future version returning null/void will throw.',
```
ADmad:
```suggestion
                    . 'In a future version returning null/void will throw an error.',
```

ADmad (commented):
Looks good once the test failure is sorted

ADmad (approved):

